PR #47, opened 2021-05-24
Include instructions on how to join Harvest slack

cgrinds:


ruanruijuan (commented):
I always wondered if there is a direct link that users can click and join. I couldn't find one.

cgrinds:
yes! I believe that is by design - an annoying design, but there nevertheless

rahulguptajss (approved):

PR #49, opened 2021-05-24
fixing slack signup link

schmots1:
I replaced the link with the short link.. www.netapp.io/slack the other link either is broken or will be in a week or two.. so use the short one.  I added this as a comment in the other issue/PR I guess you missed it.


cgrinds (approved):
Thanks David

PR #53, opened 2021-05-25
make file changes

rahulguptajss:
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,33 @@
 # Copyright 2021 NetApp, Inc.  All Rights Reserved
+.DEFAULT_GOAL:=help
 
-HARVEST_VERSION := "2.0"
+.PHONY: help deps clean build package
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,71 +1,76 @@
 # Copyright 2021 NetApp, Inc.  All Rights Reserved
+.DEFAULT_GOAL:=help
+SHELL:=/bin/bash
```
cgrinds, line 3: Not sure that path will always exist - isn't SHELL := bash sufficient?

cgrinds (commented):
Looks good

Makefile
rahulguptajss: I think we can remove it all together. Doesn't make any diff

rahulguptajss: removed it.

cgrinds (approved):

Issue #40, opened 2021-05-21
Performance metrics dont display volume name


YvosOnTheHub:
when polling performance metrics, it looks like the following:
`volume_write_latency{datacenter="lod",cluster="cluster1",node="cluster1-01",svm="cluster1-01",aggr="aggr0",type="flexvol"} 79.36234458259325`
=> volume name is missing

whereas other metrics display the following
`volume_size_available{datacenter="lod",cluster="cluster1",volume="trident_qtree_pool_nas2_LRHCOABWFT",node="cluster1-01",svm="nfs_svm",aggr="aggr1",style="flexvol"} 5367836672`
=> volume name is present

is that expected behavior?

Environment configuration:

- NetApp Lab on Demand "Using Trident with Kubernetes and ONTAP v4.0"
- host: RHEL 7.5
- installation method: yum install harvest-21.05.1-1.x86_64.rpm 
- Harvest collectors: zapi & zapiperf
- ONTAP 9.7

hashi825:
This looks to be an issue with 
https://github.com/NetApp/harvest/blob/332963c620a32c6be6e1b7b4e079b9fee50c7d77/conf/zapiperf/cdot/9.8.0/volume.yaml#L8-L10

Changing the mapping to
```yaml
 - instance_name => volume
```
Seems to fix the issue and adds the volume information back into the svm level metrics.

vgratian:
thanks for reporting this,

`instance_name` should be renamed to `volume` by the collector without the need to do it manually. I am looking to find what's the cause of the issue..

vgratian:
just recalled that I had fixed this earlier (with workload support), but we didn't to manage that branch in the GA, will open a pull request to solve this now

sridevimm:
Validated this fix in below build.
harvest version 21.06.0801-nightly (commit 5dd80cc) (build date 2021-06-08T01:08:56+0530) linux/amd64

volume_write_latency{datacenter="fas8060-39-239-241",cluster="fas8060-39-239-241",volume="container_volume_1212032_2",node="fas8060-39-239-241-02",svm="svm_1",aggr="aggr1",type="flexvol"} 0
volume_total_ops{datacenter="fas8060-39-239-241",cluster="fas8060-39-239-241",volume="container_volume_1212032_2",node="fas8060-39-239-241-02",svm="svm_1",aggr="aggr1",type="flexvol"} 18847.0228699769
[...]

PR #54, opened 2021-05-25
fix issue that instance_name not renamed/exported

vgratian:
fix issue mentioned in #40 

the ZapiPerf collector should rename the string counter `instance_name` to match its object name (e.g. `volume`). There is a bug that requires currently to do the renaming in the template, otherwise the label will be missing in the output metrics.


cgrinds (approved):
Thanks @vgratian 
Can you add a unit test to go along with this?

PR #67, opened 2021-05-26
fix issue when label is treated as a metric

### LINKED ISSUE #66: Status label on cluster_status metric disappears
**Describe the bug**
Status label on `cluster_status` metric disappears roughly after 15mins of starting Harvest

**Environment**

- Harvest version: `harvest version 21.05.1-1 (commit 2211c00) (build date 2021-05-21T01:28:12+0530) linux/amd64`
- OS: RHEL 7.9
- Install method: yum
- ONTAP Version: 9.7
- Other:

**To Reproduce**
curl -s localhost:12990/metrics | grep cluster_status

**Expected behavior**
Should report cluster_status metric as 
`cluster_status{cluster="clustername",datacenter="dc",status="ok"} 1`

**Actual behavior**
Initially reports correct metric then after 15minutes roughly the metric turns into
`cluster_status{cluster="clustername",datacenter="dc"} 1`

None

vgratian:
fix #66 

label disappears because the collector tries to store an incoming label value as a numeric metric (since both have the same name, e.g. "status" in the case reported).

currently a workaround, leaving a better solution for the future.

### DIFF
--- a/cmd/collectors/zapi/collector/zapi.go
+++ b/cmd/collectors/zapi/collector/zapi.go
@@ -286,11 +286,11 @@ func (me *Zapi) PollData() (*matrix.Matrix, error) {
 					count += 1
 				}
 			} else {
-				me.Logger.Debug().Msgf("%sskipped (%s) with value (%s): not in metric or label cache%s", color.Blue, key, value, color.End)
+				me.Logger.Debug().Msgf(" > %sskipped (%s) with value (%s): not in metric or label cache%s", color.Blue, key, value, color.End)
 				skipped += 1
 			}
 		} else {
-			logger.Trace(me.Prefix, "%sskippped (%s) with no value%s", color.Cyan, key, color.End)
+			me.Logger.Debug().Msgf(" > %sskippped (%s) with no value%s", color.Cyan, key, color.End)
 			skipped += 1
 		}
 


cmd/collectors/zapi/collector/zapi.go
```diff
@@ -269,28 +271,28 @@ func (me *Zapi) PollData() (*matrix.Matrix, error) {
...
+				logger.Trace(me.Prefix, "%slabel (%s) [%s] set value (%s)%s", color.Yellow, key, label, value, color.End)
+				count += 1
+			} else if metric := me.Matrix.GetMetric(key); metric != nil {
 				if err := metric.SetValueString(instance, value); err != nil {
-					//logger.Warn(me.Prefix, "%sskipped metric (%s) set value (%s): %v%s", color.Red, key, value, err, color.End)
+					logger.Error(me.Prefix, "%smetric (%s) set value (%s): %v%s", color.Red, key, value, err, color.End)
 					skipped += 1
 				} else {
-					//logger.Trace(me.Prefix, "%smetric (%s) set value (%s)%s", color.Green, key, value, color.End)
+					logger.Trace(me.Prefix, "%smetric (%s) set value (%s)%s", color.Green, key, value, color.End)
```
cgrinds, line 287: this was commented out before - did you mean to leave it uncommented?

cmd/collectors/zapi/collector/zapi.go
```diff
@@ -269,28 +271,28 @@ func (me *Zapi) PollData() (*matrix.Matrix, error) {
...
-				instance.SetLabel(label, value)
-				//logger.Trace(me.Prefix, "%slabel (%s) [%s] set value (%s)%s", color.Yellow, key, label, value, color.End)
-				count += 1
 			} else {
 				logger.Debug(me.Prefix, "%sskipped (%s) with value (%s): not in metric or label cache%s", color.Blue, key, value, color.End)
 				skipped += 1
 			}
 		} else {
-			//logger.Trace(me.Prefix, "%sskippped (%s) with no value%s", color.Cyan, key, color.End)
+			logger.Trace(me.Prefix, "%sskippped (%s) with no value%s", color.Cyan, key, color.End)
```
cgrinds, line 295: this was commented out before - did you mean to leave it uncommented?

cmd/collectors/zapi/collector/zapi.go
```diff
@@ -269,28 +271,28 @@ func (me *Zapi) PollData() (*matrix.Matrix, error) {
...
-		logger.Debug(me.Prefix, " > %s(%s)%s <%s%d%s> name=[%s%s%s%s] value=[%s%s%s]", color.Grey, newpath, color.End, color.Red, len(node.GetChildren()), color.End, color.Bold, color.Cyan, node.GetNameS(), color.End, color.Yellow, node.GetContentS(), color.End)
+		key := strings.Join(newpath, ".")
+		logger.Debug(me.Prefix, " > %s(%s)%s <%s%d%s> name=[%s%s%s%s] value=[%s%s%s]", color.Grey, key, color.End, color.Red, len(node.GetChildren()), color.End, color.Bold, color.Cyan, node.GetNameS(), color.End, color.Yellow, node.GetContentS(), color.End)
 
 		if value := node.GetContentS(); value != "" {
-			key := strings.Join(newpath, ".")
-			if metric := me.Matrix.GetMetric(key); metric != nil {
+			if label, has := me.instanceLabelPaths[key]; has {
+				instance.SetLabel(label, value)
+				logger.Trace(me.Prefix, "%slabel (%s) [%s] set value (%s)%s", color.Yellow, key, label, value, color.End)
```
cgrinds, line 280: did you mean to leave this uncommented

cgrinds:
Can you add a unit test that demonstrates the problem and shows this fixes it? Especially since we plan on revisiting this

cgrinds (commented):
Can you add a unit test that demonstrates the problem and shows this fixes it? Especially since we plan on revisiting this

cmd/collectors/zapi/collector/zapi.go
vgratian: no, I uncommented because it might help to understand other issues in the future

cmd/collectors/zapi/collector/zapi.go
```diff
@@ -269,28 +271,28 @@ func (me *Zapi) PollData() (*matrix.Matrix, error) {
...
 
 		newpath := append(path, node.GetNameS())
-		logger.Debug(me.Prefix, " > %s(%s)%s <%s%d%s> name=[%s%s%s%s] value=[%s%s%s]", color.Grey, newpath, color.End, color.Red, len(node.GetChildren()), color.End, color.Bold, color.Cyan, node.GetNameS(), color.End, color.Yellow, node.GetContentS(), color.End)
+		key := strings.Join(newpath, ".")
+		logger.Debug(me.Prefix, " > %s(%s)%s <%s%d%s> name=[%s%s%s%s] value=[%s%s%s]", color.Grey, key, color.End, color.Red, len(node.GetChildren()), color.End, color.Bold, color.Cyan, node.GetNameS(), color.End, color.Yellow, node.GetContentS(), color.End)
 
 		if value := node.GetContentS(); value != "" {
-			key := strings.Join(newpath, ".")
-			if metric := me.Matrix.GetMetric(key); metric != nil {
+			if label, has := me.instanceLabelPaths[key]; has {
```
rahulguptajss, line 278: instead of "has".. may be "exists" sounds better

cmd/collectors/zapi/collector/zapi.go
vgratian: > Can you add a unit test that demonstrates the problem and shows this fixes it? Especially since we plan on revisiting this

it's a bit hard, since the incoming data is from `pkg/api/ontapi/zapi`, I think we should add a method to simulate data, but this will require more time I am afraid

cmd/collectors/zapi/collector/zapi.go
vgratian: ok!

cgrinds: I would stick with `has` that's Go's way

cmd/collectors/zapi/collector/zapi.go
vgratian: what I could do is add a test for parsing the counter tree from the template, that would require less time and code changes

cgrinds: The change you made was to the `PollData` method and how it uses `instanceLabelPaths`. Can you create a test that creates a simple matrix, creates a Zapi struct, calls the `PollData` method with the matrix, then checks the relevant data after the call to `PollData`?

This method makes a zapi call - if that was refactored to a separate method seems like the relevant change around `instanceLabelPaths` could be made.

vgratian: The problem is the ZAPI call, this will fail in the test since the test unit (and the `PollData` method) will not be able to connect to ZAPI

cgrinds: That's why I was suggesting refactoring the zapi call into another method.

cgrinds (approved):

cgrinds (approved):

rahulguptajss (approved):

PR #70, opened 2021-05-26
Add harvest doctor command

cgrinds:
* Use `gopkg.in/yaml.v3` to parse yml - better parser with comment support
* Create a `HarvestConfig` struct - biggest advantage of this struct is it allows us to move away from [StringlyTyped](http://wiki.c2.com/?StringlyTyped) code
* Add `harvest doctor` cmd which:
  * Validates config file passed via `--config`
  * Prints a sanitized version of the config file: removing comments, usernames, passwords, and auth tokens - the goal is to include the output of this for every issue
* Testcase for sanitizing and struct creation

Remaining
* Evaluate the effort to replace the current config parsing code with `HarvestConfig` struct
* Add more checks to `checkAll`
* Evaluate the effort to create structs for everything in `conf/` folder. Haven't seen anything that looks problematic from a schema point of view

The goal is to merge this into main after review and continue improving

### DIFF
--- a/cmd/tools/doctor/doctor.go
+++ b/cmd/tools/doctor/doctor.go
@@ -1,0 +1,123 @@
+package doctor
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+	"goharvest2/pkg/conf"
+	"gopkg.in/yaml.v3"
+	"io/ioutil"
+	"os"
+)
+
+type options struct {
+	ShouldPrintConfig bool
+}
+
+var opts = &options{
+	ShouldPrintConfig: false,
+}
+
+var Cmd = &cobra.Command{
+	Use:   "doctor",
+	Short: "Check for potential problems",
+	Long:  "Check for potential problems",
+	Run:   doDoctorCmd,
+}
+
+func doDoctorCmd(cmd *cobra.Command, _ []string) {
+	var config = cmd.Root().PersistentFlags().Lookup("config")
+	doDoctor(config.Value.String())
+}
+
+func doDoctor(path string) {
+	contents, err := ioutil.ReadFile(path)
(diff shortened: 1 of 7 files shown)


cmd/tools/doctor/doctor.go
```diff
@@ -0,0 +1,177 @@
...
+func sanitize(nodes []*yaml.Node) {
+	for i, node := range nodes {
+		if node == nil {
+			continue
+		}
+		if node.Kind == yaml.ScalarNode && node.ShortTag() == "!!str" {
+			// Update this list of conditionals if there are other tokens you want to sanitize
+			value := node.Value
+			if value == "username" || value == "password" ||
+				value == "grafana_api_token" || value == "token" {
```
rahulguptajss, line 147: Should this be a slice?

rahulguptajss: We should redact addr also. and consult host as well.

cmd/tools/doctor/doctor.go
```diff
@@ -0,0 +1,177 @@
...
+// The downside of making all members pointers is accessing
+// the values requires more dereferencing - see doctor_test.go
+
+type Consul struct {
+	Host        *string   `yaml:"host,omitempty"`
+	ServiceName *string   `yaml:"service_name,omitempty"`
+	Tags        *[]string `yaml:"tags,omitempty"`
+}
+
+type Tools struct {
```
rahulguptajss, line 43: i think all these structs and yml parsing shall be defined in other go file.  May be only keep harvestdoctor related methods here?

cgrinds:
Addressed review comments and updated testcase to include addr and host

rahulguptajss (approved):

PR #74, opened 2021-05-26
introducing zerolog framework

rahulguptajss:
TODO: Add comments. I'll add them tomorrow and update PR.


rahulguptajss:
Hmm test case has passed locally. Looks like git issue as Chris mentioned on teams.

pkg/logging/logger.go
```diff
@@ -0,0 +1,148 @@
...
+
+var logger *Logger
+
+var once sync.Once
+
+type Logger struct {
+	*zerolog.Logger
+}
+
+func GetInstance() *Logger {
```
cgrinds, line 56: what about `Get()` instead of `GetInstance()` or `Inst()`

`GetInstance()` is kinda verbose, especially when you consider that the package already communicates logging

pkg/logging/logger.go
```diff
@@ -0,0 +1,148 @@
...
+				MaxSize:            defaultLogMaxMegaBytes,
+				MaxBackups:         defaultLogMaxBackups,
+				MaxAge:             defaultLogMaxAge}
+			logger = Configure(logConfig)
+		}
+	})
+	return logger
+}
+
+func GetInstanceSubLogger(key string, value string) *Logger {
```
cgrinds, line 77: What about Sub() or SubLogger()

cgrinds (commented):
Instead of making the same comment over and over I made the changes - take a look

cgrinds:
Seg fault when running `bin/poller -p unix --config harvest.example.yml` this happens because of this line:
`if s := me.params.GetChildContentS("log_max_bytes"); s != "" {`
at line 124.

`params` is nil since it is not read until later in the file, at line 186.

cmd/poller/poller.go
```diff
@@ -106,85 +107,93 @@ func (me *Poller) Init() error {
...
-			return err
+		logFileName = "poller_" + me.name + ".log"
+		fileLoggingEnabled = true
+	} else {
+		consoleLoggingEnabled = true
+	}
+
+	// log handling parameters
+	// size of file before rotating
+	if s := me.params.GetChildContentS("log_max_bytes"); s != "" {
```
cgrinds, line 124: seg fault since `params` is nil

cgrinds (commented):
Updated with seg fault location

cmd/collectors/unix/main.go
```diff
@@ -158,7 +158,7 @@ func (me *Unix) Init() error {
 	// load list of counters from template
 	if counters := me.Params.GetChildS("counters"); counters != nil {
 		if err = me.loadMetrics(counters); err != nil {
-			logger.Error(me.Prefix, "load metrics: %v", err)
+			me.Logger.Error().Stack().Err(err).Msgf("load metrics:")
```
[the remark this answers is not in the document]
rahulguptajss: Thanks! Agreed.

cmd/exporters/prometheus/httpd.go
```diff
@@ -23,13 +22,13 @@ func (me *Prometheus) startHttpD(addr, port string) {
 	mux.HandleFunc("/", me.ServeInfo)
 	mux.HandleFunc("/metrics", me.ServeMetrics)
 
-	logger.Debug(me.Prefix+" (httpd)", "starting server at [%s:%s]", addr, port)
+	me.Logger.Debug().Msgf(" (httpd)", "starting server at [%s:%s]", addr, port)
 	server := &http.Server{Addr: addr + ":" + port, Handler: mux}
 
 	if err := server.ListenAndServe(); err != nil {
-		logger.Fatal(me.Prefix+" (httpd)", err.Error())
+		me.Logger.Fatal().Msgf(" (httpd)", err.Error())
```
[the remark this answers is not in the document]
rahulguptajss: corrected

pkg/logging/logger.go
rahulguptajss: changed to Get()

pkg/logging/logger.go
rahulguptajss: changed to SubLogger()

cmd/poller/poller.go
rahulguptajss: Actually i did remove these in earliar commit then saw we make references to them in read me.
I am of opinion that we don't need these configuration. Probably also doesn't fit well with our docker story.
To log initial messages, i changed the log to point to Harvest.log and then point to poller logs. That doesn't seem like right approach.
Another common problem I sense if there will be too many file descriptors open for harvest.log (default config) from each poller? Let's discuss this more in meeting.

rahulguptajss: Addressed these concerns as discussed in meeting.

cgrinds (approved):
Nice job

PR #77, opened 2021-05-27
fix fat deb/rpm due to make target change

rahulguptajss:


cgrinds (approved):
Nice catch @rahulguptajss

PR #78, opened 2021-05-27
Changes in harvest.example.yml

rahulguptajss:
Currently after installing any binary we only get harvest.yml file. This file gets updated if not present in location already with harvest.example.yml. 

I am keeping harvest.example.yml in installed location for rpm/deb. This way any upgrade will have our latest template changes in harvest.yml as harvest.example.yml and customer can refer to example file. Thoughts? Do we need doc changes?

Tar gz still only ships with harvest.yml no changes there. 
For dev boxes, concept of harvest.example.yml goes away.


rpm/spec
```diff
@@ -90,6 +101,7 @@ if [ $1 = 0 ]; then
     rm /etc/systemd/system/harvest.service
     systemctl daemon-reload
     systemctl reset-failed
+    rm -rf /opt/harvest
```
cgrinds, line 104: Same comment as above

deb/postinst
```diff
@@ -14,11 +14,19 @@ chown -R harvest:harvest /var/run/harvest
 echo "  --> copying service"
 cp /opt/harvest/service/harvest.service /etc/systemd/system/
 chmod 664 /etc/systemd/system/harvest.service
-if [ ! -e /opt/harvest/harvest.yml ]; then
-    cp /opt/harvest/harvest.example.yml /opt/harvest/harvest.yml
-    echo "  --> use default config file [/opt/harvest/harvest.yml]"
+if [ -e /opt/harvest/harvest.yml ]; then
```
cgrinds, line 17: why cp harvest.yml to harvest.example.yml

deb/postinst
```diff
@@ -14,11 +14,19 @@ chown -R harvest:harvest /var/run/harvest
...
+    rm -rf /opt/harvest/harvest.example.yml
+    cp /opt/harvest/harvest.yml /opt/harvest/harvest.example.yml
+    echo "  --> copy installed harvest.yml as [/opt/harvest/harvest.example.yml]"
+fi
+if [ -e /opt/harvest/backup_harvest.yml ]; then
```
cgrinds, line 22: why the backup juggle?

deb/preinst
```diff
@@ -11,6 +11,10 @@ if [[ "$1" == "install" || "$1" == "upgrade" ]]; then
     if [  -e /usr/bin/harvest ]; then
       unlink /usr/bin/harvest && echo " --> delete link"
     fi
+    if [ -e /opt/harvest/harvest.yml ]; then
```
cgrinds, line 14: I don't think we need to backup - in the case that the file is already there - copy the new `harvest.yml` from the deb to `harvest.example.yml`. It's OK if we overwrite `harvest.example.yml`

cgrinds (commented):
I don't think the code quite matches the description above. I agree we should create a `harvest.example.yml` when rpm/deb is installed - but not much logic is needed, always write `harvest.example.yml` to the install dir. It can overwrite older `harvest.example.yml`

The only special case would be first install - in that case, the `harvest.yml` file does not exist so we create it. Every subsequent upgrade we don't touch `harvest.yml`

deb/postrm
```diff
@@ -25,5 +25,6 @@ if [[ "$1" == "purge" || "$1" == "remove" ]]; then
     if [ $(getent group harvest) ]; then
       groupdel harvest && echo " --> delete harvest group"
     fi
+    rm -rf /opt/harvest
```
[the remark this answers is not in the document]
rahulguptajss: by default rpm/deb uninstalls will remove files which they installed so harvest.yml will go away. Solution to this would be to backup yml and restore like i did for upgrade cases

deb/preinst
rahulguptajss: as per my testing, whole /opt/harvest directory gets replaced during rpm upgrade hence in pre install , i had to take backup. All of the backup stuff is to avoid overriding of files during install phase.

deb/postrm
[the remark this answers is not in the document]
vgratian: I think when $1 == "purge", then we should remove everything (maybe ask for confirmation), otherwise keep configs and caches in conventional.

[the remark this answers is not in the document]
rahulguptajss: Agree. I have made the change to keep harvest.yml and harvest.example.yml during uninstall case. For purge, no questions, everything goes away.
I haven't made any changes w.r.t conf folder as part of this story. Let's discuss on how do we handle /conf folder seperately.

deb/preinst
cgrinds: yuck, ok

rahulguptajss:
made a change where harvest yml had install user permission in deb.

cgrinds (approved):

PR #79, opened 2021-05-28
Handle special characters in passwords

cgrinds:
This change only addresses passwords in Pollers and Defaults. The bigger refactor is to use HarvestConfig through out the codebase, but that was too big a change at the moment. That change touches a lot more code.

When that change is made, the code in conf.LoadConfig can be removed.

### DIFF
--- a/pkg/conf/conf.go
+++ b/pkg/conf/conf.go
@@ -9,12 +9,58 @@ import (
 	"goharvest2/pkg/errors"
 	"goharvest2/pkg/tree"
 	"goharvest2/pkg/tree/node"
+	"gopkg.in/yaml.v3"
+	"io/ioutil"
 	"os"
 	"path"
 )
 
-func LoadConfig(config_fp string) (*node.Node, error) {
-	return tree.Import("yaml", config_fp)
+// LoadConfig loads the config info from harvest.yml
+func LoadConfig(configPath string) (*node.Node, error) {
+	configNode, err := tree.Import("yaml", configPath)
+	if configNode != nil {
+		// Load HarvestConfig to rewrite passwords - eventually all the code will be refactored to use HarvestConfig.
+		// This is needed because the current yaml parser does not handle password with special characters.
+		// E.g abc#123, that's because the # is interpreted as the beginning of a comment. The code below overwrites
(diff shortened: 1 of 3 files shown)


pkg/conf/conf.go
```diff
@@ -9,12 +9,55 @@ import (
...
+	if configNode != nil {
+		// Load HarvestConfig to rewrite - eventually all the code will be refactored to use HarvestConfig
+		_ = LoadHarvestConfig(configPath)
+		pollers := configNode.GetChildS("Pollers")
+		if pollers != nil {
+			for _, poller := range pollers.GetChildren() {
+				password := poller.GetChildContentS("password")
+				pollerStruct := (*Config.Pollers)[poller.GetNameS()]
+				if pollerStruct.Password != "" && pollerStruct.Password != password {
+					poller.SetChildContentS("password", pollerStruct.Password)
```
rahulguptajss, line 30: good to have commens here as to why we are overriding password.

pkg/conf/conf.go
```diff
@@ -9,12 +9,55 @@ import (
...
+				}
+			}
+		}
+		// Check Defaults also
+		defaultNode := configNode.GetChildS("Defaults")
+		if defaultNode != nil {
+			password := defaultNode.GetChildContentS("password")
+			defaultStruct := *Config.Defaults
+			if defaultStruct.Password != "" && defaultStruct.Password != password {
+				defaultNode.SetChildContentS("password", defaultStruct.Password)
```
rahulguptajss, line 40: same as above

vgratian:
I have a slight concern with the Poller and Exporter structs. (Sorry, this actually refers back to PR #70 which I  wasn't able to review): hardcoding poller and exporter parameters there, goes against a core architectural feature of Havest: isolation of collectors/exporters from the core modules. This is not a problem right now, since - if I understand correctly - collectors/exporters still get the untyped tree structure as their parameters, but might be a breaching the isolation principle in the future.

cgrinds:
Comment added

vgratian (commented):
> Comment added

thanks, the quote make things much more understandable

rahulguptajss (approved):

PR #88, opened 2021-05-31
add zerolog pretty print for console

rahulguptajss:


cgrinds (approved):
Looks good. 
Related, I noticed that zerolog handles escape sequences differently. If this console writer change doesn't fix that, can you remove the escape sequences that were carried over as part of the change to zerolog?

rahulguptajss:
> Looks good.
> Related, I noticed that zerolog handles escape sequences differently. If this console writer change doesn't fix that, can you remove the escape sequences that were carried over as part of the change to zerolog?

Seems to work in my dev box. 

image

Issue #63, opened 2021-05-25
InfluxDB exporter should support URL end-point


cgrinds:
**Is your feature request related to a problem? Please describe.**
From Slack:
```
Steve S  3:19 PM
for the influxdb exporter is it possible to just pass a URL?

Chris Grindstaff  3:21 PM
@Steve S can you give an example of what you want? you mean instead of decomposing the url into an addr and port?

Steve S  3:25 PM
correct - we have a large scale configuration behind load balancer  (so we’re not hitting a single host).

Chris Grindstaff  3:28 PM
make sense - I'll create an issue to track - if you're interested this is the line of code to change if you want to give it a try
```

**Describe the solution you'd like**
Allow me to exporter to a URL instead of address and IP

ruanruijuan:
The copy paste from slack didn't carry the URL:
https://github.com/NetApp/harvest/blob/fc433fecae2c445e65c9465a3166b2960552e962/cmd/exporters/influxdb/influxdb.go#L108

steverweber:
Can the code be updated so we can set the url without (bucket, org, token).
influxdb version 1.8 does not have (bucket, org, token) and a bunch of people are still on the 1.8.* release because the 2.* has big changes.

If the config accepted the following... then I assume we would get support for the older version of influxdb.
```
Exporters:
  influx:
    use_insecure_tls: true # < also perhaps nice to have
    url: https://influxdb.my:8086/write?db=mydb&u=myuser&p=mypassword&precision=s
```

steverweber:
also you would need to make the token optional. and skip setting the Authorization header for the above influxdb1.8 fix to work.

vgratian:
we would also need password/username to work with 1.8, right?

steverweber:
in influxdb1.8 you can include the username and password in the url `u= and p=` else you can pass the creds using http basic authentication.
`https://influxdb.my:8086/write?db=mydb&u=myuser&p=mypassword&precision=s`

cgrinds:
@steverweber we'll add InfluxDB `1.8` support to the list. As you noted, we only support 2.0 at the moment. Are you interested in trying to add this?

steverweber:
@cgrinds sure i'll make a PR tomorrow.

steverweber:
this is working on my end.
https://github.com/steverweber/harvest/commit/bb988c880cb9c8f0d694f24342b4e3d16c3da4c0

with config like:
```
Exporters:
  influx:
[...]

PR #93, opened 2021-05-31
Influx support url

vgratian:
to address #63, added optional parameter `url`: when it is used, `addr` will be discarded, no scheme and port will be added to construct the final URL.

also updated test and doc.


rahulguptajss (approved):

cgrinds (commented):
Nice job @vgratian 
Can you also update these?
* the struct in `pkg/conf/conf.go` to include this new key
* `testConfig.yml` to include this key
* `doctor_test.go` with an example

vgratian:
> Nice job @vgratian
> Can you also update these?
> 
> * the struct in `pkg/conf/conf.go` to include this new key
> * `testConfig.yml` to include this key
> * `doctor_test.go` with an example

done, although I am not sure if these tests are really helpful.
note, that I also moved `allow_addrs_regex` to the `prometheus` exporters, since it is not a parameter of the InfluxDB exporter.

also please note, that (commented [here](https://github.com/NetApp/harvest/pull/79#issuecomment-850216765) earlier), I do think defining component parameters in `pkg/conf` slightly breaches the architectural principle of Harvest and I think we need to discuss this separately.

rahulguptajss (approved):

cgrinds (approved):

PR #97, opened 2021-06-01
auth test cases

rahulguptajss:


.gitignore
```diff
@@ -17,4 +17,5 @@
 
 # build directories
 bin/
-dist/
\ No newline at end of file
+dist/
+cert/
```
cgrinds, line 21: was this needed? you ended up using testdata instead

rahulguptajss: yes during local testing for cert, i don't want this folder to be seen in git status.

cgrinds (approved):

PR #98, opened 2021-06-01
Replace CCLA.pdf with online link to CCLA

cgrinds:


ruanruijuan (commented):
The changes in the contributing.md look good to me.

Are we going to remove the https://github.com/NetApp/harvest/blob/main/CONTRIBUTING_CCLA.pdf file also?

PR #101, opened 2021-06-02
Zapiperf workload objects

vgratian:
fix for #9 

as mentioned in the issue, this is an inefficient way to support workloads (some counters are collected twice), leaving a better implementation for the future.

the 4 workload objects are disabled by default and the data poll interval is set to 3 minutes


vgratian:
sorry, messed up with commits, closed to open a new pr

PR #106, opened 2021-06-03
Remove buildmode=plugin

cgrinds:
Add support for cluster simulator
WIP Implement Caddy style plugins for collectors
Fix go vet warnings in node.go


Makefile
```diff
@@ -102,44 +100,6 @@ harvest: deps
 	@echo "Building grafana tool"
 	@GOOS=$(GOOS) GOARCH=$(GOARCH) go build -o bin/grafana -ldflags=$(LD_FLAGS) cmd/tools/grafana/main/main.go
 
-###############################################################################
-# Collectors
-###############################################################################
-collectors: deps
```
rahulguptajss, line 108: Thank you so much for removing for loops 🥇

cmd/collectors/zapi/collector/zapi.go
```diff
@@ -103,6 +113,25 @@ func (me *Zapi) InitVars() error {
 	return nil
 }
 
+func (me *Zapi) LoadPlugin(kind string, abc *plugin.AbstractPlugin) plugin.Plugin {
+	switch kind {
```
rahulguptajss, line 117: may be log for default case in switch?

pkg/api/ontapi/zapi/client.go
```diff
@@ -66,8 +64,11 @@ func New(config *node.Node) (*Client, error) {
 		return nil, errors.New(errors.MISSING_PARAM, "addr")
 	}
 
-	url = "https://" + addr + ":443/servlets/netapp.servlets.admin.XMLrequest_filer"
-
+	if config.GetChildContentS("is_kfs") == "true" {
```
rahulguptajss, line 67: do you want to add this variable to static structs in harvest doctor?

cmd/poller/poller.go
```diff
@@ -645,20 +654,16 @@ func (me *Poller) load_exporter(name string) exporter.Exporter {
 		return nil
 	}
 
-	binpath = path.Join(me.options.HomePath, "bin", "exporters")
-
-	if sym, err = dload.LoadFuncFromModule(binpath, strings.ToLower(class), "New"); err != nil {
-		logger.Error().Msgf("dload: %v", err.Error())
+	absExp := exporter.New(class, name, me.options, params)
```
rahulguptajss, line 657: where did we initialized class var in this method?

cmd/poller/poller.go
```diff
@@ -570,8 +557,11 @@ func (me *Poller) load_collector(class, object string) error {
 				continue
 			}
 
-			col = NewFunc(collector.New(class, object.GetNameS(), me.options, template.Copy()))
-			if err = col.Init(); err != nil {
+			col, err = me.newCollector(class, object.GetNameS(), template)
+			if col == nil {
+				continue
```
rahulguptajss, line 562: add log if it is null?

cmd/poller/poller.go
```diff
@@ -590,17 +580,19 @@ func (me *Poller) load_collector(class, object string) error {
 	logger.Debug().Msgf("initialized (%s) with %d objects", class, len(collectors))
 	// link each collector with requested exporter & update metadata
 	for _, col = range collectors {
-
+		if col == nil {
+			continue
```
rahulguptajss, line 584: add log if null?

cmd/poller/collector/collector.go
```diff
@@ -95,6 +94,13 @@ type AbstractCollector struct {
 	countMux *sync.Mutex // used for atomic access to collectCount
 }
 
+func (AbstractCollector) HarvestModule() plugin.ModuleInfo {
```
rahulguptajss, line 97: looks like a placeholder to make abtractcollector interface member. is Unix a good default?

cmd/collectors/zapi/collector/zapi.go
cgrinds: done - this  method will eventually be rewritten using modules too

cmd/poller/poller.go
cgrinds: line 652
if class = params.GetChildContentS("exporter"); class == "" {

cmd/poller/collector/collector.go
cgrinds: Good catch - this is a placeholder that can be removed since the pattern is shown in pollers

rahulguptajss (approved):
looks good. One general comment for constants. Instead of hardcoded strings, we should move them constant package in pkg folder

PR #107, opened 2021-06-03
handle panic in collector

rahulguptajss:


cmd/poller/collector/collector.go
```diff
@@ -235,7 +236,11 @@ func Init(c Collector) error {
 
 // Start will run the collector in an infinity loop
 func (me *AbstractCollector) Start(wg *sync.WaitGroup) {
-
+	defer func() {
+		if r := recover(); r != nil {
+			me.Logger.Error().Stack().Err(errors.New(errors.GO_ROUTINE_PANIC, string(debug.Stack()))).Msg("Collector panicked")
```
cgrinds, line 241: Don't we already have stacks with [zerolog](https://github.com/rs/zerolog#error-logging-with-stacktrace)?

cmd/poller/poller.go
```diff
@@ -754,21 +756,12 @@ func init() {
...
 	defer func() {
-		//logger.Warn("(main) ", "defer func here")
 		if r := recover(); r != nil {
-			logger.Info().Msgf("harvest poller panicked: %v", r)
-			// if logger still available try to write there as well
-			// do this last, since might make us panic as again
-			logger.Fatal().Msgf("(main) %v", r)
-			logger.Fatal().Msg(`(main) terminating abnormally, tip: run in foreground mode (with "--loglevel 0") to debug`)
-
+			logger.Fatal().Stack().Err(errors.New(errors.GO_ROUTINE_PANIC, string(debug.Stack()))).Msg(`(main) terminating abnormally, tip: run in foreground mode (with "--loglevel 0") to debug`)
```
cgrinds, line 764: Same comment here

rahulguptajss: recover return type is not error. One way to get stacktrace is like this.

cmd/poller/collector/collector.go
rahulguptajss: recover return type is not error. One way to get stacktrace is like this.

cgrinds (approved):

PR #113, opened 2021-06-04
enable caller for zerolog

rahulguptajss:


cgrinds:
Pasting the conversation Rahul and I had here for posterity
> enabling this has around a 2% hit on cpu

cgrinds (approved):

PR #118, opened 2021-06-06
Add unique prom ports and export type

cgrinds:
checks to doctor


cmd/tools/doctor/doctor.go
```diff
@@ -53,7 +69,110 @@ func checkAll(path string, contents []byte) {
...
+	}
+	// Add all exporters that have a port to a
+	// map of portNum -> list of names
+	seen := make(map[int][]string)
+	for name, exporter := range *config.Exporters {
+		if exporter.Port == nil || exporter.Type == nil || *exporter.Type != "Prometheus" {
+			continue
+		}
+		previous := seen[*exporter.Port]
+		previous = append(previous, name)
```
rahulguptajss, line 147: previous may be nil. It will work but why to add nil to slice.

rahulguptajss (approved):

cmd/tools/doctor/doctor.go
cgrinds: yes, previous will be nil when the map is empty
but `append` turns that nil slice into a new one
that's idiomatic Go and cleaner than if previous = nil {create new slice}

Issue #100, opened 2021-06-02
Harvest Plugins


cgrinds:
# Harvest Plugins

Harvest `v21.05.01` has limited support for dynamically linked Go plugins built using `buildmode=plugin`. Unfortunately Go's support for dynamic plugins is weak and comes with significant drawbacks. 

Basically, Go plugins were not designed as a way for other people to extend your app. They were designed for you to extend your app. 

Before outlining the pros and cons of Go's plugins, let's explore why we want them.

## Why do we want Harvest plugins?

1. Plugins allow 3rd parties to extend Harvest without (re)building it. You program to Harvest's API and, at runtime, Harvest dynamically loads your code into its process and calls it.

2. You only "pay" for what you use. If you don't use a feature from plugin A, you don't pay for it in disk footprint, you don't load the code into memory, etc. This is less important than #1.

Plugins, as a concept, are great. They allow us to build a loosely-coupled modular system. Harvest's current implementation doesn't address 1 or 2, and arguably introduces more problems than it solves.

## Cons of Go Plugins

1. Plugins and Harvest must be complied with the exact same version of Go.
2. Plugins and Harvest must be compiled with the same `GOPATH`.
3. Any imported packages between Harvest and the plugin must be the exact same version.
4. [Plugins and Harvest can NOT vendor dependencies](https://github.com/golang/go/issues/18827). If either have vendored dependencies, Harvest won't work.
5. Debuggers don't work with dynamic code - this means you can't use a debugger with Harvest right now because the interesting parts of Harvest are implemented as Go plugins.
6. Makes cross compiling harder or impossible (see Windows, Alpine, etc.)
7. Creates ~3x larger executables - with `buildmode=plugin` the `bin` directory is 140 MB, without `buildmode=plugin` the `bin` directory is 48M
8. ~7x slower builds
  
```
# with buildmode=plugin
$ make clean
$ time GOOS=darwin make build
Executed in   24.94 secs

# without buildmode=plugin
$ make clean
$ time GOOS=darwin make build
Executed in    3.39 secs
```

## Experience of others

We're not the only team to hit issues with Go plugins. I haven't found a project that recommends them.

### Traefik

**Traefik** tried and abandoned plugins due to development pain

https://github.com/traefik/traefik/issues/1336#issuecomment-474883783
https://github.com/traefik/traefik/issues/1336#issuecomment-474888814

  > [OK, bad news...](https://github.com/traefik/traefik/pull/1370#issuecomment-291903522)
[...]

PR #120, opened 2021-06-07
Module arch improvement

vgratian:
First, to safe us the language-confusion, about the terms used below:
- **plugin** - the plugins that Harvest runs as part of a collector
- **module** - a directory with several files that self-identify as a Go "package", this can be an exporter, collector, plugin or something else.

This is a follow-up on the changes applied by #106 for issue #100. This changes have greatly simplified Harvest and removed some unnecessary complications due to our use of [Go's plugin library](https://golang.org/pkg/plugin/) (not to be confused with Harvest's plugins). On the other hand, there are a number of drawbacks:

- List of collectors is hardcoded in the import section of `cmd/poller/poller.go`: if a user or developer wants to add a new collector, they have to edit that file (and re-edit after each update of Harvest).
- List of exporters is hardcoded in several sections of `cmd/poller/poller.go`
- List of plugins is hardcoded in several modules
- The gist of the new module-design is written in `cmd/plugin/plugin.go`. This module originally defined the plugin type and interfaces and implemented its base type ("AbstractPlugin"). Combining two unrelated concepts, will lead to unnecessary confusion in the future.
- Only collectors are implemented to use the plugin.RegisterModule() method

How I propose to improve:
- All Harvest modules that implement a collector, exporter or plugin are a defined in `modules.txt`. Users can safely edit this file to add new modules, including third-party modules.
- Running `make modules` will generate the file `cmd/poller/modules/modules.go` which will import all of these modules. This command will also optionally install these modules if they are not part of Harvest (e.g. third-party collectors).
- A separate module handles registering modules: `cmd/poller/registrar/registrar.go`. Each module will explicitly register either as a collector, plugin or exporter using the methods `registrar.RegisterCollector()`, `registrar.RegisterExporter()`, `registrar.RegisterPlugin()`.
- Since the arguments of these functions are fixed, no type-checking is necessary at initialization-time.
- Modules don't need to declare additional types, like `HarvestModule`, for the same reason.
- Poller will initialize these modules similar to what is done currently with collectors before but with more simplicity.

I think these improvements will give Harvest a lot of flexibility. Adding new collectors, exporters or plugins does not require to touch the core code-base of Harvest. We only need to edit `harvest.yml` and `modules.txt`. I have even tried using [this "third-party" collector](https://github.com/vgratian/GithubCollector) this way. 

Only issue left to address:
- If collector is from a third-party repository, you still have to manually copy the template to 'conf/...'. But I think this can be easily solved as well. (I didn't touch this issue, because I think we have to revisit how templates are loaded anyway).


cgrinds:
@vgratian Can you incorporate the changes we discussed in our last meeting?
- [ ] investigate using go generate instead of echoes
- [ ] investigate if there's a good way to have modules consume extra config from `harvest.yml`
- [ ] consolidate concepts, pick one of: extension, plugin, module, etc. 

I think your PR is the next natural step for what was started in #100.

cgrinds (commented):
Left general comment above

PR #128, opened 2021-06-08
stop renamed pollers

rahulguptajss:


cmd/harvest/harvest.go
```diff
@@ -319,12 +327,61 @@ func checkPollerIdentity(cmdline, pollerName string) bool {
...
+			return false
+		}
 	}
 	return true
 }
 
+func stopGhostPollers(search string, skipPoller []string) {
+	result, err := util.GetProcessPid(search)
+	if err != nil {
+		fmt.Println(err)
```
cgrinds, line 341: maybe more context here

pkg/util/util.go
```diff
@@ -50,3 +57,32 @@ func EqualStringSlice(a, b []string) bool {
...
+	var r []string
+	for _, str := range s {
+		if str != "" {
+			r = append(r, str)
+		}
+	}
+	return r
+}
+
+func GetProcessPid(search string) (string, error) {
```
cgrinds, line 80: would it be better to return a []int of pids instead of requiring the consumer to handle trimming, atoi, etc?

cgrinds: also simplifies error handling for consumer

cmd/harvest/harvest.go
```diff
@@ -319,12 +327,61 @@ func checkPollerIdentity(cmdline, pollerName string) bool {
...
+	pids := util.RemoveEmptyStrings(strings.Split(result, "\n"))
+	for _, pid := range pids {
+		p, err := strconv.Atoi(strings.TrimSpace(pid))
+		if err != nil {
+			fmt.Println(err)
+			continue
+		}
+		c, err := util.GetCmdLine(p)
+		if err != nil {
+			fmt.Println(err)
```
cgrinds, line 353: more context here?

cmd/harvest/harvest.go
```diff
@@ -319,12 +327,61 @@ func checkPollerIdentity(cmdline, pollerName string) bool {
...
+			continue
+		}
+
+		if checkPollerIdentity(c, "", false) {
+			// skip if this poller is defined in harvest config
+			var skip bool
+			for _, s := range skipPoller {
+				if strings.Contains(c, s) {
```
cgrinds, line 361: since this is a substring match, depending on naming convention in harvest.yml this may match when it shouldn't

cmd/harvest/harvest.go
```diff
@@ -319,12 +327,61 @@ func checkPollerIdentity(cmdline, pollerName string) bool {
...
+					skip = true
+					break
+				}
+			}
+			// if poller doesn't exists in harvest config
+			if !skip {
+				proc, err := os.FindProcess(p)
+				if err != nil {
+					fmt.Println(err)
```
cgrinds, line 370: more context about not finding pid we expect?

cmd/harvest/harvest.go
```diff
@@ -319,12 +327,61 @@ func checkPollerIdentity(cmdline, pollerName string) bool {
...
+					continue
+				}
+				// send terminate signal
+				if err := proc.Signal(syscall.SIGTERM); err != nil {
+					if os.IsPermission(err) {
+						fmt.Println(err)
```
cgrinds, line 376: better msg about no permission - probably matching our other no permission msg

pkg/util/util.go
```diff
@@ -50,3 +57,31 @@ func EqualStringSlice(a, b []string) bool {
...
 	return true
 }
+
+func GetCmdLine(pid int) (string, error) {
+	data, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/cmdline", pid))
+	if err != nil {
+		return "", err
+	}
+	result := string(bytes.ReplaceAll(data, []byte("\x00"), []byte(" ")))
+	return result, err
```
cgrinds, line 67: more idiomatic and clearer to return result, nil

pkg/util/util.go
```diff
@@ -50,3 +57,31 @@ func EqualStringSlice(a, b []string) bool {
 	}
 	return true
 }
+
+func GetCmdLine(pid int) (string, error) {
```
cgrinds, line 61: this is only used by `harvest.go` maybe premature to move to a util and make public

pkg/util/util.go
```diff
@@ -50,3 +57,31 @@ func EqualStringSlice(a, b []string) bool {
...
+	var r []string
+	for _, str := range s {
+		if str != "" {
+			r = append(r, str)
+		}
+	}
+	return r
+}
+
+func GetProcessPid(search string) (string, error) {
```
cgrinds, line 80: this is only used by `harvest.go` maybe premature to move to a util and make public

cmd/harvest/harvest.go
rahulguptajss: done

cmd/harvest/harvest.go
rahulguptajss: done

cmd/harvest/harvest.go
rahulguptajss: done

cmd/harvest/harvest.go
rahulguptajss: done

pkg/util/util.go
rahulguptajss: this looks like a util. I like this to be in util.

pkg/util/util.go
rahulguptajss: done

pkg/util/util.go
rahulguptajss: same as above. Looks good candidate for util

pkg/util/util.go
```diff
@@ -77,11 +81,47 @@ func RemoveEmptyStrings(s []string) []string {
...
+	fields := strings.Fields(source)
+	for _, w := range fields {
+		if w == search {
+			return true
+		}
+	}
+	return false
+}
+
+//func main() {
```
cgrinds, line 121: Remove

PR #166, opened 2021-06-14
Pass through all container arguments exactly as specified

### LINKED ISSUE #163: Docker container handles arguments strangely
**Describe the bug**

The entrypoint script takes exactly two arguments and passes these unquoted to the harvest executable. This is surprising, and makes it tricky to work out how to pass additional arguments, such as `--verbose`. It would be better if the entrypoint wrapper would pass through all arguments exactly as specified on the comm

**Examples**

The default arguments are surprising, as passing `'--config /opt/harvest.yml'` as a single string is not how harvest itself describes its options should be passed, and passing the arguments in the same way directly to harvest outside of docker throws an error.

```
docker run harvest start '--config /opt/harvest.yml'
```

The verbose flag is silently ignored

```
docker run harvest start '--config /opt/harvest.yml' '--verbose'
```

The list of pollers to run are silently ignored, and all pollers run instead

```
docker run harvest start '--config /opt/harvest.yml' poller1 poller2
```

**Expected behavior**

All arguments should be passed through exactly as specified by the user

PR to follow

optiz0r:
Previously the entrypoint wrapper took exactly two arguments, and passed them along
unquoted, to harvest meaning any spaces made the given arguments split into additional
passed arguments. This is confusing, because there's no way to know without reading
the entrypoint code why that's happening, and it's unlike any other software out
there. Any additional arguments given to the entrypoint were _silently_ ignored.
    
This commit changes the entrypoint to pass _all_ arguments through unmodified.
This is more intuitive and consistent with other applications, including the
harvest application itself.
    
**Compatibility break**
    
Where users have previously customised container arguments as:
    
```
'start' '--config /opt/harvest.yml ...'
```
    
These instead need to be passed as:

```
'start' '--config' '/opt/harvest.yml' ...
```
    
Fixes #163

### DIFF
--- a/docker/allPollerInOneContainer/Dockerfile
+++ b/docker/allPollerInOneContainer/Dockerfile
@@ -39,4 +39,4 @@ VOLUME     [ "/opt/harvest" ]
 VOLUME     [ "/var/log/harvest" ]
 
 ENTRYPOINT ["/usr/local/bin/dumb-init", "--", "./docker-entrypoint.sh"]
-CMD ["start", "--config harvest.yml"]
+CMD ["start", "--config", "harvest.yml"]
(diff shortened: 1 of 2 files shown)


rahulguptajss:

PR #167, opened 2021-06-14
Infinite sleep without needing a TTY or side effects

### LINKED ISSUE #164: Docker container stops after pollers are started when run detached/without TTY
**Describe the bug**

Trying to run the docker container via a scheduler, am seeing the container exit immediately after the pollers are spawned. This is because since the pollers are all backgrounded, there needs to be an additional process run indefinitely to keep the container alive. The method used (`sh`) only works when the container is run in attached mode, with stdin connected to the terminal. When run in detached mode (as happens when run via a scheduler such as nomad), stdin is closed, the shell process immediately exits, and thus the container stops.

Additionally, because the shell process is attached to stdin, any input given to the container is treated as shell commands and executed. This is unexpected, and potentially dangerous.

**To Reproduce**

Run container without being attached to a terminal (i.e. no `-ti` options).

**Expected behavior**

Container should keep running indefinitely

**Actual behavior**

Container exits almost immediately

Use an infinite sleep instead of spawning a shell.

PR to follow

optiz0r:
Infinite sleep without needing a TTY or side effects
    
This commit changes the means by which the container keeps itself
alive from running a `sh` process, to an explicit infinite sleep.
    
Using a shell means the container does not continue to run after
poller startup if run in detached mode, since stdin is closed the
shell process immediately exits. When run in attached mode,
use of shell will interpret any stdin as commands to be run within
the container, which is potentially dangerous, and certainly unexpected.
    
Fixes #164

### DIFF
--- a/docker/allPollerInOneContainer/docker-entrypoint.sh
+++ b/docker/allPollerInOneContainer/docker-entrypoint.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 bin/harvest $1 $2
-`exec "sh"`
+`exec /bin/sleep infinity`


rahulguptajss:

PR #168, opened 2021-06-14
Replace busybox wget with standalone package

### LINKED ISSUE #165: Cannot build container in an environment requiring a http proxy
**Describe the bug**

In an environment where direct internet access is not available, and use of an http proxy is required, the docker container cannot be built due to upstream busybox issue https://github.com/mirror/busybox/issues/21

**To Reproduce**

Build the container with `http_proxy`. `https_proxy` env vars pointing at a valid proxy

**Expected behavior**

Container image should build successfully

**Actual behavior**

The build fails where wget tries to download `dumb-init` with varying error messages

Use upstream wget instead of the wget functionality baked into busybox

optiz0r:
Docker container cannot be built in corporate environments using
http proxies due to issue https://github.com/mirror/busybox/issues/21.
    
This change installs the wget package, so a fully-featured wget
is available for downloading artifacts during container build.
    
This change could be reverted once the wget issue is resolved.
    
Fixes #165

### DIFF
--- a/docker/allPollerInOneContainer/Dockerfile
+++ b/docker/allPollerInOneContainer/Dockerfile
@@ -26,6 +26,9 @@ RUN cp -aR bin $BUILD_DIR/conf $BUILD_DIR/docs $BUILD_DIR/grafana $BUILD_DIR/doc
 
 FROM alpine:latest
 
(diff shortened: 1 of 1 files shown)


rahulguptajss:

PR #171, opened 2021-06-15
Prometheus port range

cgrinds:
@rahulguptajss I saw your branch and thought about shelving this one, but decided to go ahead and push since it was from the POC I mentioned last week. Take a look and see what you think. The yaml unmarshal makes it a little cleaner. I'll push some tests tomorrow.


rahulguptajss:
Thanks @cgrinds . These changes do not cover scenario when pollers are added to existing harvest.yml.  In this case, it's easy for this code to fail where it picks an existing used port for another poller. I added these cases as part https://github.com/NetApp/harvest/pull/175 .

PR #175, opened 2021-06-15
portrange support in harvest

rahulguptajss:


pkg/conf/conf.go
```diff
@@ -170,33 +174,101 @@ func GetHarvestLogPath() string {
...
+func GetHarvestPidPath() string {
+	var pidPath string
+	if pidPath = os.Getenv("HARVEST_PIDS"); pidPath == "" {
+		pidPath = "/var/run/harvest/"
+	}
+	return pidPath
+}
+
+func main() {
+	configFp := "/home/rahulg2/code/github/harvest/harvest.yml"
```
vgratian, line 186: forgot to clean?

pkg/conf/conf.go
```diff
@@ -170,33 +174,101 @@ func GetHarvestLogPath() string {
...
+						if k == port {
+							// remove ports which are already used
+							delete(ports, k)
+							break
+						}
+					}
+					fmt.Printf("chosen port %d \n", port)
+					return port, nil
+				} else if *exporter.Port != 0 {
+					fmt.Printf("port---- %d \n", *exporter.Port)
```
vgratian, line 234: same, I think this is for debugging

rahulguptajss: Thanks. Yes. I'll remove them before merge.

pkg/conf/conf.go
```diff
@@ -170,33 +174,101 @@ func GetHarvestLogPath() string {
...
-		}
-		for _, ec := range exportChildren {
-			promNode := definedExporters.GetChildS(ec)
-			if promNode == nil {
-				fmt.Printf("poller [%s] specified exporter [%s] that does not exist\n", p.GetNameS(), ec)
-				continue
+func GetPrometheusExporterPorts(pollerName string, configFp string) (int, error) {
+	var port int
+
+	if Config == (HarvestConfig{}) {
```
cgrinds, line 203: For the code to reach here the config should have already been loaded, perhaps added for debug?

pkg/conf/conf.go
```diff
@@ -1,6 +1,8 @@
 /*
  * Copyright NetApp Inc, 2021 All rights reserved
  */
+//package main
```
cgrinds, line 4: remove this too

pkg/conf/conf.go
```diff
@@ -170,33 +174,101 @@ func GetHarvestLogPath() string {
 	return logPath
 }
 
+func GetHarvestPidPath() string {
```
cgrinds, line 177: remove this

pkg/conf/conf.go
```diff
@@ -170,33 +174,101 @@ func GetHarvestLogPath() string {
...
+		LoadPrometheusExporterPortRangeMapping(configFp)
+	}
+	exporters := (*Config.Pollers)[pollerName].Exporters
+
+	if exporters != nil && len(*exporters) > 0 {
+		for _, e := range *exporters {
+			exporter := (*Config.Exporters)[e]
+			if *exporter.Type == "Prometheus" {
+				if exporter.PortRange != nil {
+					//fmt.Println(exporter.PortRange)
```
cgrinds, line 216: remove or add to debug?

pkg/conf/conf.go
```diff
@@ -170,33 +174,101 @@ func GetHarvestLogPath() string {
...
-				currentPort := definedExporters.GetChildS(ec).GetChildContentS("port")
-				port = currentPort
+			continue
+		}
+
+	}
+	return port, errors.New(errors.ERR_CONFIG, "No free port found for poller "+pollerName)
+}
+
+var promPortRangeMapping = make(map[string]map[int]struct{})
```
cgrinds, line 246: perhaps a struct would be clearer instead of a map of maps?

pkg/conf/conf.go
```diff
@@ -170,33 +174,101 @@ func GetHarvestLogPath() string {
...
+
+func LoadPrometheusExporterPortRangeMapping(configFp string) {
+	if Config == (HarvestConfig{}) {
```
cgrinds, line 249: For the code to reach here the config should have already been loaded, perhaps added for debug

pkg/conf/conf.go
```diff
@@ -170,33 +174,101 @@ func GetHarvestLogPath() string {
...
+		LoadHarvestConfig(configFp)
+	}
+	exporters := *Config.Exporters
+	for k, v := range exporters {
+		if *v.Type == "Prometheus" {
+			if v.PortRange != nil {
```
cgrinds, line 255: maybe pull the yaml unmarshaller from my branch so we don't need to parse here, makes this simpler

pkg/conf/conf.go
```diff
@@ -170,33 +174,101 @@ func GetHarvestLogPath() string {
...
+				portRange := v.PortRange // [2000-2030]
+				var ports = make(map[int]struct{})
+				r := regexp.MustCompile(`(\d+)\s*-\s*(\d+)`)
```
cgrinds, line 258: minor but the regex compile can be done once at yaml unmarshal time

pkg/util/util.go
```diff
@@ -146,3 +147,16 @@ func ContainsWholeWord(source string, search string) bool {
 	}
 	return false
 }
+
+func CheckPortAvailable(addr string, port string) bool {
```
cgrinds, line 151: let's try to come up with a solution that doesn't require port scanning

cgrinds (commented):
thanks @rahulguptajss 
Few minor comments and cleanup

Let's see if we can combine the best ideas from both branches. I'd like to come up with a solution that doesn't require port probing, but if we allow arbitrary harvest.yml edits that doesn't seem possible. A limited set of edits could be supported without port probing, but this design needs to be given more thought.

pkg/conf/conf.go
rahulguptajss: i made it public so thought of adding this as failsafe. I can make it private and remove the load.

pkg/conf/conf.go
rahulguptajss: it is but given its a public method and can be consumed elsewhere. Wanted to make sure it's loaded.

rahulguptajss: kept it as it is.

pkg/conf/conf.go
rahulguptajss: done

pkg/conf/conf.go
rahulguptajss: done. with [2000-2030] parsing is failing. It works fine when port range is defined as (2000-2030)

pkg/util/util.go
rahulguptajss: Is your concern around if this would make harvest start slow? Pls suggest any alternative.

pkg/conf/conf.go
rahulguptajss: map looks better. My aim was to keep look up O(1). With structs and loops, it will end up O(n)

cgrinds: a map of a map is usually yelling for more structure
This seems clearer `map[string]PortMap`
plus a bit more encapsulation. I'll make the changes and show you what I mean

cgrinds (approved):
Looks good - all the back and forth improved this one

PR #189, opened 2021-06-17
Add generate systemd subcommand

cgrinds:


pkg/conf/conf.go
```diff
@@ -51,30 +55,43 @@ func LoadConfig(configPath string) (*node.Node, error) {
 }
 
 var Config = HarvestConfig{}
+var configRead = false
 
 func LoadHarvestConfig(configPath string) error {
+	if configRead {
```
rahulguptajss, line 61: This will be false here even during first call of this method. Makes configRead false always?

cgrinds: false on first call so line 61 is false and does not return, then set to true on line 70
e.g.
```
conf.LoadHarvestConfig(path) <=== this call will load and cache config
conf.LoadHarvestConfig(path) <=== this one will return the cached config
conf.LoadHarvestConfig(path) <=== this one will return the cached config
```

rahulguptajss (approved):

Issue #132, opened 2021-06-08
Prometheus exporter: duplicate labels


vgratian:
**Describe the bug**
This was reported by a customer: some metrics have duplicate labels (e.g. `
aggr_disk_user_read_latency{datacenter="xxx",cluster="xxx",node="xxx",node="",aggr="xxx"}`

vgratian:
I was not able to reproduce this issue

vgratian:
I was able to reproduce: duplicate labels happen in the ZapiPerf collector when the same label is added under both `global_labels` and `instance_keys`. Will fix this in [this branch](https://github.com/NetApp/harvest/tree/prom-make-telegraf-compatible).

Issue #104, opened 2021-06-02
Prometheus exporter: incomplete metatags


vgratian:
**Describe the bug**
- instance labels (when exported as separate metric) and status metrics do not get the "HELP" and "TYPE" metatags

This was reported by a customer, I will add more details when I get them from the customer and/or reproduce the issues myself.

vgratian:
I moved the second issue to #132

cgrinds:
Please provide test steps @vgratian.
Here's what I tried
```
Exporters:
    prometheus2:
        exporter: Prometheus
        port: 12991
        local_http_addr: 0.0.0.0
Pollers:
    ocum-mobility-01-02:
        datacenter: nane
        addr: -REDACTED-
        auth_style: password
        username: -REDACTED-
        password: -REDACTED-
        use_insecure_tls: true
        add_meta_tags: true
        collectors:
            - Zapi
            - ZapiPerf
        exporters:
            - prometheus2
``` 

Restarted that poller, waited 10 minutes.
Executed `curl -s localhost:12990/metrics | grep "#"`
expected `HELP` lines

None appeared

PR #193, opened 2021-06-18
Prometheus exporter: make sure all parse requirements are met

vgratian:
This branch solves #132 and adds follow-up testing and fixes for #104

### 1. Need for a thorough testing & validation of metrics

There are two reasons why we need a more comprehensive testing of our metrics:
 
First, while these metrics are most commonly collected by the Prometheus DB itself, a number of alternatives can be used, such as [telegraf](https://github.com/influxdata/telegraf). Since Harvests' Prometheus exporter, follows the definitions of the open-metric format, it can be easily integrated to any other tools that support this format. The challenge however is that this tools might apply stricter requirements to the metrics. E.g. they might require HELP/TYPE tags to be always present, while Prometheus DB treats them as optional.

Secondly, simple unit tests would not catch all issues generated by our Prometheus exporter, since different plugins manipulate collected data such that exported metrics might degrade over time.

### 2. Proposed remedy: validator.py

A solution is to have a script that acts as a Prometheus scraper, collects metrics from Harvest for several intervals and prints out warnings if the metrics do not meet the strictest parsing rules. For this I wrote the [cmd/exporters/prometheus/validator.py](cmd/exporters/prometheus/validator.py). This is based on a script that was used in Harvest 2.0 Alpha - hence *Python*. We might want to move it into Go, since that will make it easier to integrate with the source/definitions [Prometheus' common](https://github.com/prometheus/common/tree/main/model) and will not require to have Python3 installed on the testing machine.

### 3. Fixes added by the PR

Using the above-mentioned script, I found and fixed several issues. Note that some this fixes would require a revision of other components of Harvest, but I left a proper fix for later. I think this is OK, since baseline usage of the Prometheus exporter (i.e when Prometheus DB scrapes Harvest), is fine without this fixes.

Overview of fixed issues:
- duplicate labels in some metrics
- - Cause: same label was used as both an instance label and as a global label (mostly in 7mode templates).
- - Solved with: whenever possible corrected in the templates. Because does not cover all cases, also added a check in the `Render()` method of the Prometheus exporter (temporary fix).

- inconsistent labels
- - Cause: metric name conflicts (i.e. same metric name was generated from two different ZapiPerf objects)
- - Solved with: renaming metrics in templates

- duplicate HELP/TYPE tags
- - Cause: this happens only for metadata that come from different collectors
- - Solved with: HTTPD will filter tags before emitting the metrics. This is slightly expensive, but only when the `add_meta_tags` options is set to true.

All these fixes, make metrics emitted  by Harvest compatible with Prometheus DB but also any other scraper that supports the open-metrics format.


cgrinds:
Thanks @vgratian - agreed that it would be nice to have the validator in Go instead of Python, but not required right now. 

Maybe we can use another projects's open-metrics endpoint validation in the future as another validator. 

Minor, but might be worth squashing your 30 commits into one before be rebase and merge.

cmd/exporters/prometheus/httpd.go
```diff
@@ -136,6 +147,32 @@ func (me *Prometheus) ServeMetrics(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func filterMetaTags(metrics [][]byte) [][]byte {
```
cgrinds, line 150: Could you add a comment or unit test with an example of a line that will be filtered because of meta tag check?

vgratian: done

> Thanks @vgratian - agreed that it would be nice to have the validator in Go instead of Python, but not required right now.
> 
> Maybe we can use another projects's open-metrics endpoint validation in the future as another validator.
> 
> Minor, but might be worth squashing your 30 commits into one before be rebase and merge.

I tried to squash the commits, but I fear I messed up somehow. Should I maybe squash the commits into a new branch and re-open a PR?

cgrinds: Yes please try that. You additional comment looks good

cgrinds (approved):

rahulguptajss (approved):

cmd/exporters/prometheus/httpd.go
vgratian: > Yes please try that. You additional comment looks good

@cgrinds, opened a new PR: #211

vgratian: please ignore last comment, we are able to squash from Github when merging

PR #200, opened 2021-06-19
Make GetCmdLine work on Darwin so that

cgrinds:
harvest status works on Mac

Marching ever closer to `shirou/gopsutil` :)


rahulguptajss:
Why not move to gopsutil citing such use cases?

cgrinds:
Still doesn't make sense for what we need - I POCed it and it didn't make a huge difference.
That's what we'll do if we continue needing to add more management functions.

PR #206, opened 2021-06-21
WIP Poller should use Harvest config

cgrinds:
struct instead of nodes
The methods suffixed with 2: `GetPollers2`, `GetPoller2`, etc. will replace the node based versions once we update all senders

Compare the two approaches to union in Poller.Union2 and Conf.Union

### DIFF
--- a/cmd/poller/poller.go
+++ b/cmd/poller/poller.go
@@ -42,6 +42,7 @@ import (
 	"goharvest2/pkg/logging"
 	"goharvest2/pkg/matrix"
 	"goharvest2/pkg/tree/node"
+	"gopkg.in/yaml.v3"
 	"net/http"
 	_ "net/http/pprof" // #nosec since pprof is off by default
 	"os"
@@ -90,7 +91,7 @@ type Poller struct {
 	collectors     []collector.Collector
 	exporters      []exporter.Exporter
 	exporterParams *node.Node
-	params         *node.Node
+	params         *conf.Poller
 	metadata       *matrix.Matrix
 	status         *matrix.Matrix
 }
@@ -117,7 +118,7 @@ func (p *Poller) Init() error {
 		consoleLoggingEnabled = true
 	}
 
-	if p.params, err = conf.GetPoller(p.options.Config, p.name); err != nil {
+	if p.params, err = conf.GetPoller2(p.options.Config, p.name); err != nil {
 		// separate logger is not yet configured as it depends on setting logMaxMegaBytes, logMaxFiles later
 		// Using default instance of logger which logs below error to harvest.log
 		logging.SubLogger("Poller", p.name).Error().Stack().Err(err).Msg("read config")
@@ -126,17 +127,13 @@ func (p *Poller) Init() error {
 
 	// log handling parameters
 	// size of file before rotating
-	if s := p.params.GetChildContentS("log_max_bytes"); s != "" {
-		if i, err := strconv.ParseInt(s, 10, 64); err == nil {
-			logMaxMegaBytes = int(i / (1024 * 1024))
-		}
+	if s := p.params.LogMaxBytes; s != nil {
+		logMaxMegaBytes = int(*s / (1024 * 1024))
 	}
 
 	// maximum number of rotated files to keep
-	if s := p.params.GetChildContentS("log_max_files"); s != "" {
-		if i, err := strconv.Atoi(s); err == nil {
-			logMaxBackups = i
-		}
+	if s := p.params.LogMaxFiles; s != nil {
+		logMaxBackups = *p.params.LogMaxFiles
 	}
 
 	logConfig := logging.LogConfig{ConsoleLoggingEnabled: consoleLoggingEnabled,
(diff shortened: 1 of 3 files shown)


cmd/poller/poller.go
```diff
@@ -575,6 +576,39 @@ func (me *Poller) loadCollector(class, object string) error {
 	return nil
 }
 
+func Union2(template *node.Node, params *conf.Poller) {
```
rahulguptajss, line 579: Could you add some examples or comments in this method. Not very clear to follow.

pkg/conf/conf.go
```diff
@@ -338,6 +368,52 @@ type Poller struct {
 	Exporters      *[]string `yaml:"exporters,omitempty"`
 	Collectors     *[]string `yaml:"collectors,omitempty"`
 	IsKfs          *bool     `yaml:"is_kfs,omitempty"`
+	PollerSchedule *string   `yaml:"poller_schedule,omitempty"`
+}
+
```
Hardikl, line 373: comments with one small example.

cmd/poller/poller.go
```diff
@@ -186,26 +183,30 @@ func (me *Poller) Init() error {
...
-		fp := ""
-		for i := range filenames {
-			if fp = me.params.GetChildContentS(filenames[i]); fp == "" {
-				// use default paths
-				// example: /opt/harvest/cert/hostname.key, /opt/harvest/cert/hostname.pem
-				fp = path.Join(me.options.HomePath, "cert/", me.options.Hostname+extensions[i])
-				me.params.SetChildContentS(filenames[i], fp)
-				logger.Debug().Msgf("using default [%s] path: [%s]", filenames[i], fp)
+	if me.params.AuthStyle != nil && *me.params.AuthStyle == "certificate_auth" {
+		if me.params.SslCert != nil {
```
Hardikl, line 194: i think it should be == nil

cmd/poller/poller.go
```diff
@@ -575,6 +576,39 @@ func (me *Poller) loadCollector(class, object string) error {
...
+					switch valNode.Tag {
+					case "!!str", "!!bool":
+						newNode.Content = []byte(valNode.Value)
+					case "!!seq":
+						for _, seqNode := range valNode.Content {
+							newNode.NewChildS(seqNode.Value, seqNode.Value)
+						}
+					}
+					template.AddChild(newNode)
+				}
```
Hardikl, line 606: Yeah, not able to follow the logic here. pls add comment with example.

Hardikl, line 606: Yeah, not able to follow the logic here. pls add comment with example.

pkg/conf/conf.go
```diff
@@ -338,6 +368,52 @@ type Poller struct {
...
+func (p *Poller) Union(defaults *Poller) {
+	if p.Datacenter == nil && defaults.Datacenter != nil {
+		p.Datacenter = defaults.Datacenter
+	}
```
Hardikl, line 377: comments with one small example.

cgrinds:
Added comment to `Union2` take a look and let me know if that's clearer. Also keep in mind I'm showing two different ways of achieving the same thing so we can discuss the best approach. That's why the original PR said 
> Compare the two approaches to union in `Poller.Union2` and `Conf.Union` 

The gist of the comparison is `Union2` uses reflection which makes it kinda icky, slower, and somewhat opaque/magical. The big advantage is when a field is added/removed from the `Poller` struct, the `Union2` method doesn't have to change. 

`Union` on the other hand is easy to understand, no magic, and very explicit, but whenever a field is added to the `Poller` struct, you have to remember to update the `Union` method or it will fail to do the right thing at runtime, sometimes silently (field renames and removals the compiler will catch).

The other notable difference between the two methods is one is comparing a node to a poller, while the other is comparing a poller to a poller. That's not that big a deal though. Reflection or explicit are viable options in both cases.

Go prides itself on being boring and obvious so the more idiomatic is probably `Union`. I'm fine with `Union` except for the keeping sync problem, that makes me lean towards `Union2`.

cmd/poller/poller.go
```diff
@@ -575,6 +576,56 @@ func (me *Poller) loadCollector(class, object string) error {
...
+// from poller to hNode.
+// If the field already exists in hNode, nothing is copied.
+// Instead of comparing each field of the poller individually and being forced
+// to keep this method in sync with the Poller struct, reflection via yaml marshaling
+// is used to do the comparison. First the poller is marshaled to yaml and then
+// unmarshalled into a list of generic yaml node. Each generic yaml node is walked, checking
+// if there is a corresponding node in hNode, when there isn't one, a new hNode is created
+// and populated with the yaml node's content. Finally, the new hNode is added to its parent
+
+func Union2(hNode *node.Node, poller *conf.Poller) {
```
rahulguptajss, line 591: Thanks comments are very helpful. I agree reflection brings better maintenance.
Are we only considering implementing struct till poller level or zapi also?
If we want to replace all then this union method isn't needed.

cgrinds: Struct for all config related things initially.  Maybe zapis later.
ZAPI is a much bigger challenge - best would be to take zapidoc and generate structs, similar to KFS.
There are some big downsides with that approach though, one of the biggest being you end up with a struct per ONTAP version.

Hardikl (dismissed):
changes look good.

PR #208, opened 2021-06-21
Add Harvest CUE validation and type-checking

cgrinds:
Validate and eval with:
```
cue vet harvest.cue harvest.yml
cue eval -s harvest.cue harvest.yml
```


vgratian (approved):
would be good to add short info about using cue in the documentation or wiki (maybe as an additional troubleshooting step).

PR #211, opened 2021-06-22
Prom make telegraf compatible 2

vgratian:
Re-opened PR. For description please see #193. Only difference is that testing is added for `prometheus.FilterMetaTags()`.


vgratian:
please ignore, I was able to squash the previous PR

PR #216, opened 2021-06-23
fix collector lagging log

rahulguptajss:


cmd/poller/collector/collector.go
```diff
@@ -382,7 +382,7 @@ func (me *AbstractCollector) Start(wg *sync.WaitGroup) {
 		if nd := me.Schedule.NextDue(); nd > 0 {
 			me.Logger.Debug().Msgf("sleeping %s until next poll", nd.String()) //DEBUG
 			me.Schedule.Sleep()
-		} else if nd.Milliseconds() > -50 { // avoid warning for small delays
+		} else if nd.Milliseconds() > -30000 { // log if lagging by more than 30 secs
```
Hardikl, line 385: one more else might be needed to differentiate more than 30s and less than 30s.

cgrinds (approved):

PR #218, opened 2021-06-23
port range test cases and harvest doctor validation

rahulguptajss:


pkg/conf/conf_test.go
```diff
@@ -0,0 +1,41 @@
...
+		test{"failure_no_free_port", args{pollerName: "cluster-01"}, true},
+		test{"failure_poller_name_does_not_exist", args{pollerName: "test1"}, true},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := GetPrometheusExporterPorts(tt.args.pollerName)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("GetPrometheusExporterPorts() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
```
Hardikl, line 35: not clear about this, if wantErr is true, then it would show error if actual error comes ?

Hardikl (changes requested):
minor comment for error logging

pkg/conf/testdata/testConfig.yml
```diff
@@ -0,0 +1,51 @@
+
```
cgrinds, line 1: Might be better to update the current testdata/testConfig.yml in doctor instead of creating another. Two config files is more work to keep in sync than one and the only change with this one is port_range or refactor both to use the same

pkg/conf/conf_test.go
rahulguptajss: no if we get error then test cases pass

pkg/conf/testdata/testConfig.yml
rahulguptajss: done

cgrinds (approved):

Hardikl (approved):

PR #225, opened 2021-06-23
Environ check does not work on Unix when running as non-root user

cgrinds:


rahulguptajss (approved):

rahulguptajss:
needs merging in main

PR #245, opened 2021-06-25
tar gz support docker

rahulguptajss:


Makefile
```diff
@@ -109,8 +109,8 @@ dist-tar:
 	@rm -rf ${DIST}
 	@mkdir ${TMP}
 	@mkdir ${DIST}
-	@cp -a bin conf docs grafana README.md LICENSE ${TMP}
-	@cp -a harvest.yml ${TMP}/harvest.yml
+	@cp -r .git cmd bin conf docker docs grafana pkg service go.mod go.sum Makefile README.md LICENSE ${TMP}
```
cgrinds, line 112: Including `.git` is risky - the tar.gz doesn't need to be able to build harvest just run it. Don't think these are needed to run:
- .git
- pkg
- go.mod
- go.sum
- Makefile

rahulguptajss: we need .git for getting commit id for binaries built inside docker image. Others are needed as well for building docker image.

cgrinds: ok so all the additions are needed to enable building docker images - the other way to avoid this would be to include a dockerfile that's used to build harvest. If we did that, we wouldn't need to include these extras.

Let's include these files until we have the publish workflow down
Thanks @rahulguptajss

cgrinds (approved):

Issue #249, opened 2021-06-25
Unix collector unable to read /proc filesystem


cgrinds:
This happens on RHEL and Deb systems due to changes in https://github.com/NetApp/harvest/pull/141/ that set the SUID bit for executables in `harvest/bin` so they always run as the `harvest` user. 

The `harvest` user does not have permission to read the `/proc/` filesystem which means the Unix collector fails to start.

One workaround is to run as root directly or via sudo

cgrinds:
Log msg
```
Jun 25 19:10:22 scspo2403645002 harvest[19127]: 7:10PM ERR goharvest2/cmd/collectors/unix/main.go:167 > load metrics error="read file => io: open /proc/19132/io: permission denied" Poller=unix1 collector=Unix:poller stack=[{"func":"New","line":"35","source":"errors.go"},{"func":"(*Process).loadIo","line":"267","so>
```

ruanruijuan:
The customer who took the 21.05.3 pre-release build hit this issue too.

vgratian:
I suspect the problem is that in `deb/postinst`:

```
chmod -R u+s /opt/harvest/bin
chmod -R g+s /opt/harvest/bin
```
changes the set-user-ID to root (current executing user), not to harvest which is what we wanted.

rahulguptajss:
Tested in ubuntu, works fine when harvest is started through systemctl. It fails and reports same error as above when started through bin/harvest which we don't recommend on rpm/deb

setuid looks fine.

![image](https://user-images.githubusercontent.com/25551691/123644427-000a1000-d843-11eb-8a0a-3778701df386.png)

vgratian:
can you check to user ID of the poller when started from CLI vs. when started from systemd?

vgratian:
could you check if the user-ID of the poller is harvest as well when you start them with systemctl?

rahulguptajss:
yes it is harvest in both cases.

vgratian:
I reproduced this issue on a Debian 8 with harvest version `21.08.3011-v21.05.0 (commit 19f8f25)`. I log in as root and when I start a poller, it has the following unix credentials:

* Real User-ID: `root`
* Real Group-ID: `root`
* Effective User-ID: `harvest`
* Effective Group-ID: `harvest`

[...]

PR #250, opened 2021-06-25
Update changelog

cgrinds:


ruanruijuan (approved):
Is there a workaround to document for #249 ?

cgrinds:
> Is there a workaround to document for #249 ?

Yes, several possible workarounds, but none documented in detail yet: change permissions, run as different user, change mount for `proc` so `harvest` user can access it

rahulguptajss (approved):

PR #258, opened 2021-06-26
refactor: replace brittle union with mergo

cgrinds:


pkg/conf/conf.go
```diff
@@ -381,48 +382,7 @@ type Poller struct {
...
-	if p.Collectors == nil && defaults.Collectors != nil {
-		p.Collectors = defaults.Collectors
-	}
-	if p.IsKfs == nil && defaults.IsKfs != nil {
-		p.IsKfs = defaults.IsKfs
-	}
-	if p.PollerSchedule == nil && defaults.PollerSchedule != nil {
-		p.PollerSchedule = defaults.PollerSchedule
-	}
+	_ = mergo.Merge(p, defaults)
```
rahulguptajss, line 385: nice. Means no more reflection code needed as well?

cgrinds: More accurate to say we moved the reflection from our code to theirs 😄
but that's still a win. We also still need to keep `Union2` until we remove node from config code.

rahulguptajss (approved):

PR #260, opened 2021-06-27
test: ensure flow style collector/exporters work

cgrinds:


cmd/tools/doctor/testdata/testConfig.yml
```diff
@@ -113,4 +113,8 @@ Pollers:
   sim-0001:
     is_kfs: true
 
+  flow:
```
rahulguptajss, line 116: Is this a new tag?

cgrinds: no - line 116 is just the name of the poller. The flow style of yaml is just another way to specify list/maps that our earlier yaml parser did not support. Flow style just means a more compact inline form.
This
`exporters: [prometheus1, prom2, prom3]`
instead of
```
exporters:
  - prometheus1
  - prom2
  - prom3
```

It's nice when you have one exporter or poller especially.
This PR only adds tests for the flow style since we have supported every since we moved to yaml3.

rahulguptajss: got it

rahulguptajss (dismissed):

PR #261, opened 2021-06-27
Better validation

cgrinds:


pkg/conf/conf_test.go
```diff
@@ -96,3 +96,29 @@ func TestPollerStructDefaults(t *testing.T) {
 		}
 	})
 }
+
+func TestFlowStyle(t *testing.T) {
```
rahulguptajss, line 100: same code @ https://github.com/NetApp/harvest/pull/260/files#diff-e9a8689643f96da6e2b4f53665ce69ed7456e62e70c7b9a9df81f5efe6bd0ed5R100 also?

cgrinds: thanks - removed

rahulguptajss (dismissed):

PR #266, opened 2021-06-28
feat: Struct change for config

Hardikl:


pkg/conf/conf.go
```diff
@@ -215,7 +230,7 @@ func GetHarvestHomePath() string {
 func GetHarvestLogPath() string {
 	var logPath string
 	if logPath = os.Getenv("HARVEST_LOGS"); logPath == "" {
-		logPath = "/var/log/harvest/"
+		logPath = "/Users/hardikl/Documents/HarvestPublic/harvest/log/harvest/"
```
cgrinds, line 233: Whoops. Make sure you change this back. I assume you changed this because you're on a Mac? If so, ping me, and I'll show you a better solution using the env var above

pkg/conf/conf.go
```diff
@@ -444,6 +459,7 @@ type Exporter struct {
 	Token         *string `yaml:"token,omitempty"`
 	Precision     *string `yaml:"precision,omitempty"`
 	ClientTimeout *string `yaml:"client_timeout,omitempty"`
+	Version       *string `yaml:"version,omitempty"`
```
cgrinds, line 462: Nice catch! Make sure you include `version` in one of the tests too.

cmd/exporters/prometheus/prometheus.go
```diff
@@ -81,10 +81,10 @@ func (me *Prometheus) Init() error {
...
 	}
 
-	if x := me.Params.GetChildContentS("global_prefix"); x != "" {
-		me.Logger.Debug().Msgf("will use global prefix [%s]", x)
-		me.globalPrefix = x
-		if !strings.HasSuffix(x, "_") {
+	if x := me.Params.GlobalPrefix; x != nil {
+		me.Logger.Debug().Msgf("will use global prefix [%s]", *x)
+		me.globalPrefix = *x
+		if !strings.HasSuffix(*x, "_") {
```
cgrinds, line 87: minor but instead of de-referencing again you could say
if !strings.HasSuffix(me.globalPrefix, "_") {
since you just de`referenced the line before

cmd/exporters/prometheus/prometheus.go
```diff
@@ -97,17 +97,17 @@ func (me *Prometheus) Init() error {
 	}
 
 	// add HELP and TYPE tags to exported metrics if requested
-	if me.Params.GetChildContentS("add_meta_tags") == "true" {
+	if me.Params.ShouldAddMetaTags != nil && *me.Params.ShouldAddMetaTags == true {
```
cgrinds, line 100: clearer to say
if me.Params.ShouldAddMetaTags != nil && *me.Params.ShouldAddMetaTags {

cmd/exporters/prometheus/prometheus.go
```diff
@@ -176,9 +175,9 @@ func (me *Prometheus) Init() error {
 	}
 
 	addr := localHttpAddr
-	if x := me.Params.GetChildContentS("local_http_addr"); x != "" {
-		addr = x
-		me.Logger.Debug().Msgf("using custom local addr [%s]", x)
+	if x := me.Params.LocalHttpAddr; x != nil {
+		addr = *x
+		me.Logger.Debug().Msgf("using custom local addr [%s]", *x)
```
cgrinds, line 180: minor, could say
me.Logger.Debug().Msgf("using custom local addr [%s]", addr)

cmd/exporters/influxdb/influxdb.go
```diff
@@ -121,7 +125,7 @@ func (e *InfluxDB) Init() error {
 	// construct HTTP client
 	e.client = &http.Client{Timeout: timeout}
 
-	e.Logger.Debug().Msgf("initialized exporter, ready to emit to [%s:%s]", addr, port)
+	//e.Logger.Debug().Msgf("initialized exporter, ready to emit to [%s:%s]", *addr, *port)
```
cgrinds, line 128: why comment out?

cmd/exporters/influxdb/influxdb.go
```diff
@@ -51,66 +51,70 @@ func (e *InfluxDB) Init() error {
...
 
-		url = "http://" + addr + ":" + port
-		e.url = fmt.Sprintf("%s/api/v%s/write?org=%s&bucket=%s&precision=%s", url, v, org, bucket, p)
+		urlToUSe := "http://" + *addr + ":" + strconv.Itoa(*port)
+		url = &urlToUSe
+		e.url = fmt.Sprintf("%s/api/v%s/write?org=%s&bucket=%s&precision=%s", *url, *v, *org, *bucket, *p)
 	} else {
-		e.url = url
+		e.url = *url
+		// [TODO] need to store addr and port even when url has provided.
```
cgrinds, line 108: did you mean to handle this?

cmd/exporters/influxdb/influxdb.go
```diff
@@ -51,66 +51,70 @@ func (e *InfluxDB) Init() error {
...
+	if org = e.Params.Org; org == nil {
 		return errors.New(errors.MISSING_PARAM, "org")
 	}
-	e.Logger.Debug().Msgf("using organization [%s]", org)
+	e.Logger.Debug().Msgf("using organization [%s]", *org)
 
-	if e.token = e.Params.GetChildContentS("token"); e.token == "" {
+	if token = e.Params.Token; token == nil {
 		return errors.New(errors.MISSING_PARAM, "token")
 	} else {
```
cgrinds, line 72: else block not needed since if does a return

cmd/exporters/influxdb/influxdb_test.go
```diff
@@ -6,27 +6,36 @@ package influxdb
...
 
 // test that the addr (and port) parameters
 // are handled properly to construct server URL
 func TestAddrParameter(t *testing.T) {
-
 	expectedURL := "http://localhost:8086/api/v2/write?org=netapp&bucket=harvest&precision=s"
 
 	opts := &options.Options{}
 	opts.Debug = true
+	var params map[string]conf.Exporter
```
cgrinds, line 21: minor but somewhat consistent throughout this PR - it would be better to call these exporters - it made sense to call it params earlier with the node code, but it's clearer to call it exporters now

cmd/exporters/influxdb/influxdb_test.go
```diff
@@ -6,27 +6,36 @@ package influxdb
...
+	exporterName := "influx-test-addr"
 
-	params := node.NewS("")
-	params.NewChildS("addr", "localhost")
-	params.NewChildS("org", "netapp")
-	params.NewChildS("bucket", "harvest")
-	params.NewChildS("token", "xxxxxxx")
+	path := "../../tools/doctor/testdata/testConfig.yml"
+	err := conf.LoadHarvestConfig(path)
```
cgrinds, line 25: not needed since you're calling  `conf.GetExporters2` below

cmd/exporters/influxdb/influxdb_test.go
```diff
@@ -6,27 +6,36 @@ package influxdb
...
+	if err != nil {
+		panic(err)
+	}
+
+	if params, err = conf.GetExporters2(path); err != nil {
```
cgrinds, line 30: like above, not needed

cgrinds, line 30: like above, redundant

cgrinds, line 30: not needed

cgrinds, line 30: not needed

cgrinds, line 30: not needed

cgrinds, line 30: not needed

cgrinds (changes requested):
Nice job!

cmd/exporters/influxdb/influxdb.go
Hardikl: actually in else block it's assigning token to e.token, which would be used later

cmd/exporters/influxdb/influxdb.go
Hardikl: addr and port only be assigned and initialized when url is null, else they are null. And line 128 prints both of them which gives NPE. 

So, either we can parse url and update addr and port accordingly when url is present, else change the logger at 128.

cmd/exporters/influxdb/influxdb.go
Hardikl: mentioned above due to NPE in one of the test cases when url is not null.

cmd/exporters/influxdb/influxdb_test.go
Hardikl: done

cmd/exporters/influxdb/influxdb_test.go
Hardikl: done

cmd/exporters/prometheus/prometheus.go
Hardikl: changed.

pkg/conf/conf.go
Hardikl: Yups.
Reverting it.

pkg/conf/conf.go
Hardikl: Added the test.

cmd/exporters/influxdb/influxdb.go
Hardikl: handled this.

cmd/exporters/influxdb/influxdb.go
Hardikl: handled this.

cmd/exporters/influxdb/influxdb.go
Hardikl: handled this.

cmd/exporters/influxdb/influxdb.go
Hardikl: handled this.

cmd/exporters/influxdb/influxdb.go
cgrinds: What I meant was the return in the `if` means you're code can be made more clear like this:
```
if token = e.Params.Token; token == nil {
    return errors.New(errors.MISSING_PARAM, "token")
}
e.token = *token
e.Logger.Debug().Msg("will use authorization with api token")
```

cmd/exporters/influxdb/influxdb_test.go
```diff
@@ -41,18 +46,55 @@ func TestAddrParameter(t *testing.T) {
...
 func TestUrlParameter(t *testing.T) {
+	expectedURL := "https://some-valid-domain-name.net:8888/api/v2/write?org=netapp&bucket=harvest&precision=s"
+
+	opts := &options.Options{}
+	opts.Debug = true
+	var exporters map[string]conf.Exporter
+	exporterName := "influx-test-url"
+	var err error
+
+	path := "../../tools/doctor/testdata/testConfig.yml"
```
cgrinds, line 57: I should have mentioned this yesterday, but this block of seven or so lines, that is repeated in all the tests, could be refactored to a single method they all call.

cmd/tools/doctor/testdata/testConfig.yml
```diff
@@ -36,6 +36,27 @@ Exporters:
     token: abcdefghijklmnopqrstuvwxyz
   invalid-type:
     exporter: Foo
+  influx-test-addr:
```
cgrinds, line 39: Nice refactoring out of the code 👍

pkg/conf/conf.go
```diff
@@ -88,6 +88,21 @@ func SafeConfig(n *node.Node, fp string) error {
...
 }
 
+func GetExporters2(configFp string) (map[string]Exporter, error) {
+	err := LoadHarvestConfig(configFp)
+	if err != nil {
+		return nil, err
+	}
+	exporters := Config.Exporters
+
+	if exporters == nil {
```
rahulguptajss, line 98: why is it an error if exporter section not found? It should just be nil right?

cmd/exporters/influxdb/influxdb.go
```diff
@@ -51,66 +52,82 @@ func (e *InfluxDB) Init() error {
...
+		   step 2: localhost:8088 value in addrAndPort var
+		   step 3: addr has localhost and port has 8088
+		*/
+		addrAndPort := strings.Split(strings.Split(e.url, "//")[1], "/")[0]
+		res := strings.Split(addrAndPort, ":")
+		addr = &res[0]
+		if p, err := strconv.Atoi(res[1]); err != nil {
+			e.Logger.Warn().Msgf("invalid port [%s]", p)
+		} else {
+			port = &p
```
rahulguptajss, line 120: do you need port just for log message later? If yes, why not just log whole url and avoid all this parsing.

rahulguptajss (changes requested):

cmd/exporters/influxdb/influxdb.go
Hardikl: ohh, I understood different way. Got it.

cmd/exporters/influxdb/influxdb.go
Hardikl: Agreed. Parsing may not be full proof and url would be more better. 
Url is already logged, so not needed to log again.

cmd/exporters/influxdb/influxdb_test.go
Hardikl: Yes. Done.

pkg/conf/conf.go
Hardikl: actually as part of this PR, just changed node to struct, logic wise it's same as earlier as GetExporters.

cmd/exporters/influxdb/influxdb.go
```diff
@@ -51,66 +52,82 @@ func (e *InfluxDB) Init() error {
...
 		}
 
-		if port = e.Params.GetChildContentS("port"); port == "" {
-			e.Logger.Debug().Msgf("using default port [%s]", defaultPort)
-			port = defaultPort
-		} else if _, err = strconv.Atoi(port); err != nil {
-			return errors.New(errors.INVALID_PARAM, "port")
+		if port = e.Params.Port; port == nil {
+			e.Logger.Debug().Msgf("using default port [%d]", defaultPort)
+			defPort := defaultPort
```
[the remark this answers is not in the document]
Hardikl: not able to access the address of constant defaultPort, so using this approach.

PR #269, opened 2021-06-28
fix exporter config

rahulguptajss:


pkg/conf/conf.go
```diff
@@ -336,8 +336,11 @@ func GetUniqueExporters(p *node.Node, configFp string) ([]string, error) {
 		}
 		exporterMap := make(map[string]string)
 		for _, ec := range exportChildren {
-			exporterType := definedExporters.GetChildS(ec).GetChildContentS("exporter")
-			exporterMap[exporterType] = ec
+			e := definedExporters.GetChildS(ec)
+			if e != nil {
+				exporterType := definedExporters.GetChildS(ec).GetChildContentS("exporter")
```
cgrinds, line 341: better to use `e` so
`e.GetChildContentS("exporter")`

cgrinds (approved):

PR #270, opened 2021-06-29
feat: add simple collector that works on all platforms

cgrinds:


cmd/collectors/simple/simple.go
```diff
@@ -0,0 +1,152 @@
...
+	n.Logger.Debug().Msgf("added %d, removed %d, total instances %d", a, r, t)
+
+	return nil, nil
+}
+
+// PollData - update data cache
+func (n *NodeMon) PollData() (*matrix.Matrix, error) {
+	n.Matrix.Reset()
+
+	toQuery := []string{"alloc", "num_gc", "num_cpu"}
```
rahulguptajss, line 119: template has more counters. Those are not collected?

cgrinds: yep not yet - doesn't much matter so I'll remove them

rahulguptajss (approved):

PR #281, opened 2021-06-30
docs(collector): add copyright to simple collector

cgrinds:


cmd/collectors/simple/simple.go
```diff
@@ -1,3 +1,5 @@
+// Copyright NetApp Inc, 2021 All rights reserved
```
rahulguptajss, line 1: do we need to keep on updating year as well in existing files?

rahulguptajss (approved):

cmd/collectors/simple/simple.go
cgrinds: good question - i'd prefer not

PR #282, opened 2021-07-01
feat(config): changing exporter params into struct way approach

Hardikl:


cgrinds (approved):
Nice job @Hardikl appreciate the persistence on this one 👍

PR #288, opened 2021-07-02
fix(collector): simple collector should include pid in metrics

cgrinds:


cmd/collectors/simple/simple.go
```diff
@@ -104,6 +106,7 @@ func (n *NodeMon) PollInstance() (*matrix.Matrix, error) {
 		}
 		instance.SetLabel("poller", name)
 		instance.SetLabel("version", version.VERSION)
+		instance.SetLabel("pid", strconv.Itoa(os.Getpid()))
```
rahulguptajss, line 109: does it work when started as root user (bin/harvest start) in rpms?

rahulguptajss (approved):

cmd/collectors/simple/simple.go
cgrinds: it should since it's Go's code, not the ps code

PR #293, opened 2021-07-11
feat: detect and fix ONTAP swagger errors

cgrinds:


cmd/tools/rest/swag.go
```diff
@@ -311,6 +318,65 @@ func readSwagger(args Args) (ontap, error) {
...
+	case map[string]interface{}:
+		for k, val := range vv {
+			switch k {
+			case "maxLength", "minLength", "maximum", "minimum", "minItems", "maxItems":
+				if s, ok := val.(string); ok {
+					num, err := strconv.Atoi(s)
+					if err == nil {
+						vv[k] = num
+					}
+				}
```
Hardikl, line 369: may be we should use some array at line 363 and check with contains here as this list may grow in future.

Hardikl (approved):

rahulguptajss (commented):
seems to have same files as https://github.com/NetApp/harvest/pull/292. may be merge both?

cgrinds:
> seems to have same files as #292. may be merge both?

yeah this branch was created from `cbg-rest-tool` since the auto-fix changes only make sense on top of `cbg-rest-tool`

Issue #328, opened 2021-07-21
release 21.05.4: harvest kills unrelated process


vgratian:
**Describe the bug**
Running "harvest start" kills other, unrelated processes. Seems like any process that has "poller" in its cmdline is killed.

**Environment**
My OS is Manjaro Linux, my Harvest installation is compiled locally with branches pulled from github. Bug occurs with both `main` and `release 21.05.4`. I reproduced the same issue in a Debian 8 VM, running a `.deb` installation of Harvest 21.05.4.

**To Reproduce**
In one terminal open any program that has "poller" either in its name or in its arguments. I have tried these:

```bash
$ vim cmd/poller/poller.go
```

and

```bash
$ tail -f log/poller_unix.log
```

In a second terminal start harvest poller(s):

```bash
$ bin/harvest start unix
```

The program opened in the first terminal will catch termination signal and exit. E.g.:

```bash
$ vim cmd/poller/poller.go
Vim: Caught deadly signal TERM
Vim: Finished.


Terminated
```

**Possible solution, workaround, fix**
Before starting any poller, `bin/harvest` doublechecks if it's not already running. If it is running, `bin/harvest` will kill it. Seems like we are misidentifying some other processes as harvest pollers.

rahulguptajss:
Verfied. Works fine as expected.

PR #329, opened 2021-07-21
fix: harvest start|restart should not kill processes with poller in name

cgrinds:
Fixes: #328


rahulguptajss (approved):
Needs to be merged in main as well

PR #333, opened 2021-07-22
fix: false negatives by prefix matcher

### LINKED ISSUE #332: release 21.05.4: grafana tool: prefix not added to some metrics
**Describe the bug**
New feature of the grafana tool adding prefixes to metric names (see PR #264) misses several metrics. I found two:

- in harvest_dashboard_metadata.json:
  - `count (count by (datacenter) (poller_status))` (no prefix was added)
- in harvest_dashboard_cluster.json:
  - `label_values(node_uptime{datacenter="$Datacenter"},cluster)` (no prefix was added)

**To Reproduce**
Configure your Prometheus exporter to add a global prefix to metrics, e.g.:

```yaml
  prom0:
    exporter: Prometheus
    port: 5550
    global_prefix: netapp
```

Import the dashboards with the grafana tool, adding a global prefix, e.g.:

```bash
$ harvest grafana import -f "21.05.4" -p netapp
```

In Grafana check if all graphs are displayed. 

Another way of reproducing is to add the two queries above to the test cases in `cmd/tools/grafana/grafana_test.go`.

Improve the metric-catching regex rules in the grafana tool. Although for the long term this is probably not a great solution, since there might be always corner cases that we can't catch.

vgratian:
fixes #332

For the first query, I couldn't find a better solution than just change the query in the dashboard. (Trying to match this with another regex, seems likely to produce false positives).

For the second query, a more precise regexp is used. Tests are updated as well.

### DIFF
--- a/cmd/tools/grafana/grafana.go
+++ b/cmd/tools/grafana/grafana.go
@@ -354,7 +354,7 @@ func addPrefixToMetricNames(expr, prefix string) string {
 
 	// variable queries
 	if strings.HasPrefix(expr, "label_values(") {
-		if isMatch, err = regexp.MatchString(`^label_values\(([a-zA-Z_])+({.+?})?,\s?[a-zA-Z_]+\)$`, expr); err != nil {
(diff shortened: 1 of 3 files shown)


Hardikl (approved):
looks fine.

cgrinds (approved):

PR #349, opened 2021-07-25
feat: install Grafana, Prometheus, Harvest

cgrinds:
- Added `docker generate full` command to stand-up Prometheus, Grafana, and a container per poller
- Instead of duplicating the dashboards, I moved them and updated references
    - Verfied that `bin/grafana import` works with new directory structure
- Removed allPollerInOneContainer
- Updated dashboards to include datasource template for auto-install (provisioning)


docker/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+Bring everything up :tada:
+   
+```
+docker compose -f docker/prom-stack.yml -f harvest-compose.yml up -d --remove-orphans
+```
+
+## Config and Using
+
+The `prom-stack.yml` compose file creates a `frontend` and `backend` network. Prometheus and Grafana publish their admin ports on the front-end network and are routable to the local machine. By default, the Harvest pollers are part of the backend network and do not expose their Prometheus web end-points. If you want their end-points exposed, pass the `--ports` flag to `generate` like so:
```
rahulguptajss, line 70: should be `--port` instead of `--ports`

docker/README.md
```diff
@@ -0,0 +1,106 @@
...
+### Prometheus
+
+- [List of targets](http://localhost:9090/targets)
+
+### Managing
+
+#### Stop all containers
+
+```
+docker compose -f docker/prom-stack.yml -f harvest-compose.yml down
```
rahulguptajss, line 96: docker-compose

docker/onePollerPerContainer/docker-compose.tmpl
```diff
@@ -1,23 +1,25 @@
-version: "3.5"
```
rahulguptajss, line 1: If i recall kompose (k8 artifacts) were not generated without 3+ version. pls do check once.

rahulguptajss (changes requested):

docker/README.md
```diff
@@ -0,0 +1,106 @@
...
+`generate docker full` does two things:
+1. Creates a Docker compose file with a container for each Harvest poller defined in your `harvest.yml`
+2. Creates a matching Prometheus service discovery file for each Harvest poller (located in `docker/prometheus/harvest_targets.yml`). Prometheus uses this file to scrape the Harvest pollers. 
+
+## Start everything
+
+Bring everything up :tada:
+   
+```
+docker compose -f docker/prom-stack.yml -f harvest-compose.yml up -d --remove-orphans
```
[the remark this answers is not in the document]
cgrinds: Ah this is a platform difference - compose was rolled into docker for Mac and Windows, but not Linux. The plan is to replace docker-compose everywhere but not yet. Good catch.

cmd/tools/generate/generate.go
```diff
@@ -61,40 +77,83 @@ func doDockerCompose(cmd *cobra.Command, _ []string) {
...
 	for _, v := range conf.Config.PollersOrdered {
 		port, _ := conf.GetPrometheusExporterPorts(v)
-		pollerTemplate.Pollers = append(pollerTemplate.Pollers, PollerPort{v, port, absPath, opts.loglevel, opts.image, version.VERSION})
+		pollerInfo := PollerPort{
+			PollerName:    v,
+			ConfigFile:    path,
+			Port:          port,
+			LogLevel:      opts.loglevel,
+			Image:         opts.image,
+			ContainerName: "poller_" + v,
```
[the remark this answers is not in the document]
cgrinds: it's somewhat handy, but it doesn't work in dev since the version is 2.0.2

docker/README.md
cgrinds: nice catch!

docker/onePollerPerContainer/docker-compose.tmpl
cgrinds: Yeah it's a bit of a [mess](https://docs.docker.com/compose/compose-file/compose-versioning/#versioning). Sticking with `3.5` seems reasonable.  The reason I wanted to remove it doesn't exist anymore so I'll add it back

cgrinds:
- re-added version to container name
- improved volume support so it works on all platforms (sudo docker on Linux did not work with PWD)
- replaced beta `docker compose` with `docker-compose`
- moved `prom-stack.yml` from docker directory to root so host path resolution work correctly on Linux
- rebased to incorporate Hardik's dashboard fixes

docker/onePollerPerContainer/docker-compose.tmpl
```diff
@@ -1,23 +1,27 @@
...
+      - {{ .Port }}:{{ .Port }}
+    {{- end}}
+    {{- end}}
+    command: '--poller {{ .PollerName }} {{if .Port }}--promPort {{ .Port }} {{ end }}
+    {{- if ne .LogLevel 2 }}--loglevel {{ .LogLevel }} {{ end}}--config /opt/harvest.yml'
     volumes:
       - harvest:/opt/harvest/
       - type: bind
-        source: {{ .ConfigFile }}
+        source: ./{{ .ConfigFile }}
```
rahulguptajss, line 19: Why relative path? We want absolute path here.

cgrinds: [Docker](https://stackoverflow.com/questions/42248198/how-to-mount-a-single-file-in-a-volume/59554543#59554543) turns this into an absolute path - verified on linux and macos

rahulguptajss (approved):

Issue #357, opened 2021-07-28
Harvest panics when port_range is used in the Prometheus exporter


vgratian:
**Describe the bug**
When you configure a Prometheus exporter with the `port_range` parameter, Harvest will panic. Additionally, if the exporter is added in the Defaults (rather than the Pollers) section, no free port is allocated to the exporter. See details below.

**To Reproduce**
Define a Prometheus exporter in your harvest.yml and add to your unix poller:

```yaml
Exporters:
  prom:
    exporter: Prometheus
    port_range: 5555-5559
    global_prefix: netapp
Pollers:
  unix:
    addr: localhost
    collectors:
      - Unix
     exporters:
      - prom
```

Try to start the poller:

```bash
$ ./bin/harvest start unix -ft --config harvest_.yml 

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x728efb]

goroutine 1 [running]:
goharvest2/pkg/conf.loadPrometheusExporterPortRangeMapping()
	goharvest2/pkg/conf/conf.go:328 +0x21b
goharvest2/pkg/conf.GetPrometheusExporterPorts(0xc0001aba58, 0x4, 0x4, 0xc0001aba58, 0x4)
	goharvest2/pkg/conf/conf.go:254 +0x8b4
main.getPollerPrometheusPort(0xc000155440, 0xcdb0e0, 0x4)
	command-line-arguments/harvest.go:544 +0x77
main.doManageCmd(0xc000183900, 0xc00017c6c0, 0x1, 0x4)
	command-line-arguments/harvest.go:161 +0x6ff
github.com/spf13/cobra.(*Command).execute(0xc000183900, 0xc00017c680, 0x4, 0x4, 0xc000183900, 0xc00017c680)
	github.com/spf13/cobra@v1.2.1/command.go:860 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0xcdf9c0, 0x0, 0xc943d8, 0xc00008e058)
	github.com/spf13/cobra@v1.2.1/command.go:974 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
	github.com/spf13/cobra@v1.2.1/command.go:902
main.main()
	command-line-arguments/harvest.go:661 +0x34

```
[...]

PR #358, opened 2021-07-28
fix: Prometheus port_range-related issue 357

vgratian:
This fixes issue #357

The underlying problems are:
- in [pkg/conf/conf.go:  loadPrometheusExporterPortRangeMapping()](https://github.com/NetApp/harvest/blob/664a8265788dedb4850c5f4ba3aa514005f5e11d/pkg/conf/conf.go#L316) we deference the `addr` parameter, which is not documented in the list of parameters of the Prometheus exporter. But it's also redundant, since we only care about free ports on the localhost.
- in pkg/conf/conf.go: GetPrometheusExporterPorts() we do not check for exporters in the Defaults section.

This PR fixes both issues.

### DIFF
--- a/pkg/conf/conf.go
+++ b/pkg/conf/conf.go
@@ -256,6 +256,11 @@ func GetPrometheusExporterPorts(pollerName string) (int, error) {
 	if poller == nil {
 		return 0, errors.New(errors.ERR_CONFIG, "Poller does not exist "+pollerName)
 	}
+
+	if Config.Defaults != nil {
+		poller.Union(Config.Defaults) // exporter can be defined in Defaults
+	}
+
 	exporters := poller.Exporters
 	if exporters != nil && len(*exporters) > 0 {
 		for _, e := range *exporters {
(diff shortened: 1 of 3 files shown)


pkg/conf/conf.go
```diff
@@ -256,6 +256,11 @@ func GetPrometheusExporterPorts(pollerName string) (int, error) {
 	if poller == nil {
 		return 0, errors.New(errors.ERR_CONFIG, "Poller does not exist "+pollerName)
 	}
+
+	if Config.Defaults != nil {
```
cgrinds, line 260: unioning should have already happened

I bet `harvest.go` wasn't converted to use the struct based in this codepath

cgrinds: yep that's the case - what if we fix this more globally (so we don't have more examples of this in downstream code) by adding these lines at 173 in conf.go

```
// Load struct based config here so the poller structs are union-ed with defaults
if _, err = GetPollers2(configFp); err != nil {
    return nil, err
}
```

vgratian: I has the same idea first, thanks

vgratian: I tried to use this more globally, changed cmd/harvest/harvest.go to use the struct-based Poller config

vgratian:
Closing this, since my last commits were failing commitlint, please see: #370

PR #360, opened 2021-07-29
feat: disk dashboard changes

Hardikl:
Grafana: http://10.249.25.52:3000/ admin/admin

### DIFF
--- a/conf/zapiperf/cdot/9.8.0/disk.yaml
+++ b/conf/zapiperf/cdot/9.8.0/disk.yaml
@@ -36,7 +36,7 @@ plugins:
   # any names after the object names will be threated as
   # label names that will be added to instances
     - node
-    - aggr node
+    - aggr ...
   #  - plex node aggr
   #  - raid node aggr plex
 
--- a/conf/zapi/cdot/9.8.0/disk.yaml
+++ b/conf/zapi/cdot/9.8.0/disk.yaml
@@ -6,7 +6,7 @@ object:           disk
 counters:
   storage-disk-info:
     - ^^disk-uid
-    - ^^disk-name               => disk
+    - ^^disk-name              => disk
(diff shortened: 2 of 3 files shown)


conf/zapiperf/cdot/9.8.0/disk.yaml
```diff
@@ -36,7 +36,7 @@ plugins:
   # any names after the object names will be threated as
   # label names that will be added to instances
     - node
-    - aggr node
+    - aggr ...
```
cgrinds, line 39: is ... correct?

cgrinds (commented):
nice job @Hardikl

conf/zapiperf/cdot/9.8.0/disk.yaml
Hardikl: Yes. 
So we are aggregating data at node and aggr level and all metrics like user_read_chain and disk_busy would be available at node/aggr not at disk. Also talked with Vachgan and get to know that getting at disk level may impact performance as more calculations would be involved.

But we would need info of plex and raid where we can show how many plex,raid would be in system, so metric calculation wise it's same just with this aggr_xxx_xxx metric would be showing other labels as well along with node,cluster,datacenter.

Ex: for raid count:                    
count( count by(raid) (aggr_disk_busy{datacenter="$Datacenter",cluster="$Cluster",node=~"$Node"}) )

Hardikl: just to add, flashcache and flashpool charts are pending as my cluster don't have them, doing in next PR.

cgrinds: thanks - got it, so you're wanting to copy all labels of the original instance

cgrinds (approved):

PR #370, opened 2021-08-02
fix: Prometheus port_range-related issue 357 (Re-opened PR)

vgratian:
Second PR to fix issue #357

Changed to things:
* refactored `cmd/harvest/harvest.go` to use the struct-based poller configs (this solves Poller parameters not being merged with Defaults)
* in `conf.loadPrometheusExporterPortRangeMapping()` use `localhost` when checking for free ports

(Re-opened PR since the commit messages were failing the commitlint-check)

### DIFF
--- a/cmd/harvest/harvest.go
+++ b/cmd/harvest/harvest.go
@@ -28,7 +28,6 @@ import (
 	"goharvest2/cmd/tools/zapi"
 	"goharvest2/pkg/conf"
 	"goharvest2/pkg/set"
-	"goharvest2/pkg/tree/node"
 	"goharvest2/pkg/util"
 	"net"
 	_ "net/http/pprof" // #nosec since pprof is off by default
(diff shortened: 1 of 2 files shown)


Hardikl (approved):

cmd/harvest/harvest.go
```diff
@@ -106,8 +113,7 @@ func doManageCmd(cmd *cobra.Command, args []string) {
 
 	//cmd.DebugFlags()  // uncomment to print flags
 
-	pollers, err := conf.GetPollers(opts.config)
-	if err != nil {
+	if pollers, err = conf.GetPollers2(opts.config); err != nil {
```
cgrinds, line 116: do we want the pollers returned in the same order that they're listed in the harvest.yml? I think we do for deterministic printing. If so, we can use 
`conf.Config.PollersOrdered`

cgrinds (approved):

PR #371, opened 2021-08-02
fix: HARVEST_CONF should be consulted when loading files from conf

### LINKED ISSUE #368: HARVEST_CONF not used when looking for collectors
**Describe the bug**
When running poller from a random directory while `HARVEST_CONF` is set correcty finds main `harvest.yml` config file, but fails to find dependent configurations like collectors 👍 
```
nabox-harvest2        | 12:44PM INF command-line-arguments/poller.go:153 > log level used: info Poller=cluster1
nabox-harvest2        | 12:44PM INF command-line-arguments/poller.go:154 > options config: /conf/harvest.yml Poller=cluster1
nabox-harvest2        | 12:44PM INF command-line-arguments/poller.go:179 > started in foreground [pid=24] Poller=cluster1
nabox-harvest2        | 12:44PM ERR command-line-arguments/poller.go:250 > load collector (Zapi): error="open conf/zapi/default.yaml: no such file or directory" Poller=cluster1
nabox-harvest2        | 12:44PM ERR command-line-arguments/poller.go:250 > load collector (ZapiPerf): error="open conf/zapiperf/default.yaml: no such file or directory" Poller=cluster1
nabox-harvest2        | 12:44PM WRN command-line-arguments/poller.go:257 > no collectors initialized, stopping Poller=cluster1
nabox-harvest2        | 12:44PM INF command-line-arguments/poller.go:421 > cleaning up and stopping [pid=24] Poller=cluster1
```

**Environment**

- Harvest version: harvest version 21.05.4-2 (commit 19f8f25) (build date 2021-07-22T15:42:14+0000) linux/amd64
- Command line arguments used: `/netapp-harvest/bin/poller -p cluster1`
- OS: Debian 10.18 container base image
- Install method: native
- ONTAP Version: 9.7
- Other:

**To Reproduce**
- Install Harvest in `/somewhere`
- Move to a different directory
- `export HARVEST_CONF=/somewhere`
- run harvest

**Expected behavior**
Harvest should start and find collectors

**Actual behavior**
```
error="open conf/zapi/default.yaml: no such file or directory"
```

Add `HARVEST_CONF` to search path when looking for additional yaml files

cgrinds:
directory.

Fixes #368

This change addresses the problem when using `bin/poller` but does not work with `bin/harvest`.

The issue with `bin/harvest` is the `os.Exec`-ed executables are not found without more directory work.

The workaround is `cd` into the installation directory before executing `bin/harvest`

### DIFF
--- a/cmd/harvest/harvest.go
+++ b/cmd/harvest/harvest.go
@@ -573,7 +573,7 @@ func init() {
 	rootCmd.AddCommand(manageCmd("stop", true))
 	rootCmd.AddCommand(manageCmd("restart", true))
 	rootCmd.AddCommand(manageCmd("kill", true))
-	rootCmd.AddCommand(config.ConfigCmd, zapi.Cmd, rest.Cmd, grafana.GrafanaCmd, stub.NewCmd)
+	rootCmd.AddCommand(config.ConfigCmd, zapi.Cmd, rest.Cmd, grafana.Cmd, stub.NewCmd)
 	rootCmd.AddCommand(generate.Cmd)
 	rootCmd.AddCommand(doctor.Cmd)
 	rootCmd.AddCommand(version.Cmd())
--- a/cmd/tools/grafana/grafana.go
+++ b/cmd/tools/grafana/grafana.go
@@ -70,6 +70,7 @@ func doExport(_ *cobra.Command, _ []string) {
 }
 
 func doImport(_ *cobra.Command, _ []string) {
+	opts.command = "import"
 	adjustOptions()
 	askForToken()
 	var doesFolderExist = doesGrafanaFolderExist()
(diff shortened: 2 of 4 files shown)


ybizeul:
I also found out the `grafana` has the same problem, not use `HARVEST_CONF`, should it be fixed in the same merge ?

cmd/poller/options/options.go
```diff
@@ -65,6 +65,13 @@ func SetPathsAndHostname(args *Options) {
 	}
 
 	args.HomePath = conf.GetHarvestHomePath()
+	configPath := os.Getenv("HARVEST_CONF")
+	if configPath != "" {
+		if !strings.HasSuffix(configPath, "/") {
+			configPath += "/"
+		}
+		args.HomePath = configPath
+	}
```
Hardikl, line 74: can we move this logic to conf.go file in GetHarvestHomePath function ?

cgrinds: i considered that initially, but was concerned about how frequently that method is called. After profiling it though, it's fine since most of these happen via initialization. I like it, thanks! I'll move this to `GetHarvestHomePath`

Also fixed the `grafana` problem Yann mentioned with this PR

Hardikl: As it's moved to conf.go, remove this from here as well.

Hardikl (approved):

cmd/poller/options/options.go
cgrinds: oh dear :) good catch

PR #373, opened 2021-08-02
fix: template version selection should try earliest

cgrinds:


cmd/poller/collector/helpers.go
```diff
@@ -115,9 +115,13 @@ func (c *AbstractCollector) ImportSubTemplate(model, filename string, ver [3]int
...
-// returns -1 if not match else returns equals or closest match to the left
+//getClosestIndex returns the closest left match to the sorted list of input versions
+// returns -1 when the versions list is empty
+// returns equal or closest match to the left
 func getClosestIndex(versions []*version.Version, version *version.Version) int {
+	if len(versions) == 0 {
+		return -1
+	}
 	var l = 0
 	var r = len(versions) - 1
```
cdurai-netapp, line 126: changes looks good. may be we can move var r = len(versions) before if loop .

cdurai-netapp (approved):

Hardikl (approved):

PR #378, opened 2021-08-03
docs: remove references to deprecated addr

cgrinds:
Prometheus `addr` is deprecated. Use [local_http_addr](https://github.com/NetApp/harvest/tree/main/cmd/exporters/prometheus#parameters) instead or omit entirely if you want to use `0.0.0.0`.

### DIFF
--- a/harvest.cue
+++ b/harvest.cue
@@ -1,20 +1,14 @@
 package harvest
 
-Exporters: [Name=_]: #Prom | #Influx | #PromConsul
+Exporters: [Name=_]: #Prom | #Influx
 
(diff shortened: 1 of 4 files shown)


harvest.cue
```diff
@@ -3,20 +3,14 @@ package harvest
 Exporters: [Name=_]: #Prom | #Influx | #PromConsul
```
Hardikl, line 3: remove PromConsul from here as well.

Hardikl (approved):

Hardikl (approved):

PR #392, opened 2021-08-05
feat: asup supporting changes part-1(public)

Hardikl:
TODO:
1. schedule testing yet to test from poller.go. --> Done
2. payload is partially populated, need to check  --> Done. Svm is pending.

populated payload:
{
 "Target": {
  "Version": "9.8.0",
  "Model": "cdot",
  "Serial": "1-80-000011",
  "Ping": 0
 },
 "Nodes": {
  "Count": 1,
  "DataPoints": 16,
  "PollTime": 111673,
  "ApiTime": 111416,
  "ParseTime": 138,
  "PluginTime": 9
 },
 "Volumes": {
  "Count": 97,
  "DataPoints": 3310,
  "PollTime": 366289,
  "ApiTime": 178908,
  "ParseTime": 15450,
  "PluginTime": 600
 },
 "Svms": {
  "Count": 0,
  "DataPoints": 0,
  "PollTime": 0,
  "ApiTime": 0,
  "ParseTime": 0,
  "PluginTime": 0
 },
 "Platform": {
  "OS": "",
  "Arch": "",
  "MemoryKb": 0,
  "CPUs": 0
 },
 "Harvest": {
  "UUID": "e8c343f4-6d42-4136-a866-ed756c3a10a9",
  "Version": "21.08.0619",
  "Release": "v21.05.4",
  "Commit": "8b37086",
  "BuildDate": "2021-08-06T19:15:16+0530",
  "NumClusters": 3
 }
}


Makefile
```diff
@@ -87,7 +87,7 @@ harvest: deps
 
 	@# Build the harvest poller
 	@echo "Building poller"
-	@GOOS=$(GOOS) GOARCH=$(GOARCH) go build -trimpath -o bin/poller -ldflags=$(LD_FLAGS) cmd/poller/poller.go
+	@cd cmd/poller; GOOS=$(GOOS) GOARCH=$(GOARCH) go build -trimpath -o ../../bin/poller -ldflags=$(LD_FLAGS)
```
cgrinds, line 90: this isn't needed is it?

cmd/poller/poller.go
```diff
@@ -65,6 +66,8 @@ var (
 	logMaxMegaBytes = logging.DefaultLogMaxMegaBytes // 10MB
 	logMaxBackups   = logging.DefaultLogMaxBackups
 	logMaxAge       = logging.DefaultLogMaxAge
+	asupInterval    = "1m" // send every 1
```
cgrinds, line 69: does this send every 1 minute or every 1 month?

cmd/poller/poller.go
```diff
@@ -281,6 +284,27 @@ func (p *Poller) Init() error {
 	}
 	logger.Debug().Msgf("set poller schedule with %s frequency", pollerSchedule)
 
+	// ASUP-related params will be loaded from Tools section, since this feature will
```
cgrinds, line 287: might be clearer to say
ASUP parameters will be loaded from Tools section

cmd/poller/poller.go
```diff
@@ -281,6 +284,27 @@ func (p *Poller) Init() error {
...
+	// probably be delegated to a separate tool
+	toolsParams, err := conf.GetTools(p.options.Config)
+	if err != nil {
+		logger.Error().Stack().Err(err).Msgf("load Tools config")
```
cgrinds, line 291: better to use `Msg` when there is no substitution

cmd/poller/poller.go
```diff
@@ -281,6 +284,27 @@ func (p *Poller) Init() error {
...
+		return err
+	}
+
+	logger.Info().Msgf("Tools param %s", toolsParams)
+	//if toolsParams.AsupEnabled != nil && *toolsParams.AsupEnabled {
```
cgrinds, line 296: WIP?

cmd/poller/poller.go
```diff
@@ -412,6 +437,12 @@ func (p *Poller) Run() {
 			upExporters = upe
 		}
 
+		logger.Info().Msgf("A")
```
cgrinds, line 440: reminder to remove

cmd/poller/asup.go
```diff
@@ -0,0 +1,269 @@
...
+
+	return nil, nil
+
+}
+
+// TODO: This function will be used to invoke harvest-asup(private repo)
+func (p *Poller) sendAsupMessage(msg *asupMessage) error {
+	logger.Info().Msgf("ASUP generatation started")
+
+	file, _ := json.MarshalIndent(msg, "", " ")
```
cgrinds, line 80: more efficient to write directly to the file instead of holding the json in a string and writing the string. On the other hand, these json messages are small

something like this:
```
file, err := os.OpenFile("payload.json", os.O_CREATE, os.ModePerm)
defer file.Close()
encoder := json.NewEncoder(file) 
encoder.Encode(msg)
```

cmd/poller/asup.go
```diff
@@ -0,0 +1,269 @@
...
+	// info about ONTAP host and instances
+	msg.Target, msg.Nodes, msg.Volumes = p.getInstanceInfo()
+
+	return msg, nil
+}
+
+func (p *Poller) getInstanceInfo() (*targetInfo, *instanceInfo, *instanceInfo) {
+	target := new(targetInfo)
+	nodes := new(instanceInfo)
+	vols := new(instanceInfo)
```
cgrinds, line 126: should add # of SVMs too

cmd/poller/asup.go
```diff
@@ -0,0 +1,269 @@
...
+
+func getNumClusters(configFp string) uint8 {
+	var (
+		count                       uint8
+		pollers, poller, collectors *node.Node
+		collectorName               string
+		err                         error
+	)
+
+	if pollers, err = conf.GetPollers(configFp); err == nil {
```
cgrinds, line 255: Use struct based instead of node based

Makefile
Hardikl: It's needed, As now we have 2 files in cmd/poller to compile: poller.go and asup.go.

cmd/poller/poller.go
```diff
@@ -743,6 +774,18 @@ var pollerCmd = &cobra.Command{
 	Run:   startPoller,
 }
 
+// Returns true if at least one collector is known
+// to collect from an Ontap system (needs to be updated
+// when we add other Ontap collectors, e.g. REST)
+func (p *Poller) targetIsOntap() bool {
```
cgrinds, line 780: is this method used?

cmd/poller/poller.go
Hardikl: Schedule testing is in-progress.

cmd/poller/poller.go
Hardikl: Yes, working on it.

cmd/poller/poller.go
Hardikl: yeah :)

Makefile
cgrinds: this breaks `go run cmd/poller/poller.go -p foo` let's move `asup.go` to `cmd/tools/asup` and then revert the change at line 90

cmd/poller/asup.go
```diff
@@ -0,0 +1,269 @@
...
+	vols := new(instanceInfo)
+
+	// get ping value from poller metadata
+	target.Ping, _ = p.status.LazyGetValueFloat64("ping", "host")
+
+	// scan collectors
+
+	for _, c := range p.collectors {
+
+		if c.GetName() == "ZapiPerf" {
```
cgrinds, line 135: should this be `Zapi` instead of `ZapiPerf`

cmd/poller/asup.go
```diff
@@ -0,0 +1,269 @@
...
+
+			if c.GetObject() == "SystemNode" {
```
cgrinds, line 137: should this be `Node` instead of `SystemNode`?

cmd/poller/asup.go
Hardikl: done

cmd/poller/asup.go
Hardikl: changed.

cmd/poller/asup.go
Hardikl: added

cmd/poller/asup.go
Hardikl: done. added encoder with indentation and O_WRONLY.

cmd/poller/poller.go
Hardikl: 1m would send for every minute, actual interval will be decided later.

cmd/poller/asup.go
Hardikl: added svm but not able to fetch data for svms via zapi, looking into it

cmd/poller/poller.go
Hardikl: yes, now it's used.

cmd/poller/poller.go
Hardikl: Done.

Makefile
Hardikl: Done

pkg/conf/conf.go
```diff
@@ -92,6 +92,24 @@ func LoadHarvestConfig(configPath string) error {
 	return nil
 }
 
+func SafeHarvestConfig(configPath string) error {
```
rahulguptajss, line 95: can u add some comments for this method? Also some unit test cases

pkg/conf/conf.go
```diff
@@ -92,6 +92,24 @@ func LoadHarvestConfig(configPath string) error {
...
+		err      error
+		fileName string
+	)
+	if !configRead {
+		panic("config not read")
+	}
+	if contents, err = yaml.Marshal(&Config); err != nil {
+		return err
+	}
+	if filepaths := strings.Split(configPath, "/"); len(filepaths) > 1 {
```
rahulguptajss, line 107: config path can have more than 2 slashes. In that case filename would be wrong?

pkg/conf/conf.go
```diff
@@ -92,6 +92,24 @@ func LoadHarvestConfig(configPath string) error {
...
+		fileName = filepaths[1]
+	}
+	return ioutil.WriteFile("new_"+fileName, contents, 0644)
```
rahulguptajss, line 110: What happens when multiple poller try to execute this same line? This code will have race condition.

cmd/tools/asup/asup.go
```diff
@@ -0,0 +1,294 @@
...
+
+	var (
+		arch, countString, line string
+		fields                  []string
+		count                   uint64
+		output                  []byte
+		err                     error
+	)
+
+	if output, err = exec.Command("lscpu").Output(); err == nil {
```
rahulguptajss, line 202: better to consider using https://github.com/shirou/gopsutil

rahulguptajss (changes requested):

cmd/poller/poller.go
```diff
@@ -65,6 +67,8 @@ var (
 	logMaxMegaBytes = logging.DefaultLogMaxMegaBytes // 10MB
 	logMaxBackups   = logging.DefaultLogMaxBackups
 	logMaxAge       = logging.DefaultLogMaxAge
+	asupInterval    = "1m" // send every 1 minute
```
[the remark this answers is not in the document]
cgrinds: it's probably OK if it's not configurable since it will typically be on or off, but it should be >1m. I think this is still in the testing phase

pkg/conf/conf.go
cgrinds: good catch @rahulguptajss - better to use path.Base or path.Split

pkg/conf/conf.go
cgrinds: @Hardikl and I discussed this earlier and I don't think his code reflects what we discussed yet - the idea is each poller will create a payload file prefixed with the poller name, which is unique per harvest.yml

cmd/tools/asup/asup.go
cgrinds: +1 I mentioned the same for asup payload limit checks

pkg/conf/conf.go
rahulguptajss: @Hardikl  where are we using this file again? I see that poller will still read from default config which won't have uuid?

Hardikl:
Payload detail
{
 "Target": {
  "Version": "9.10.1",
  "Model": "cdot",
  "Serial": "1-80-000011",
  "Ping": 0,
  "ClusterUuid": "3ccba2b0-f8f7-11eb-be9f-005056a70771"
 },
 "Nodes": {
  "Count": 2,
  "DataPoints": 34,
  "PollTime": 404349,
  "ApiTime": 401553,
  "ParseTime": 1544,
  "PluginTime": 52,
  "NodeUuid": [
   "921135d7-f8f5-11eb-be51-005056a70771",
   "93f16547-f8f5-11eb-8876-005056a79562"
  ]
 },
 "Volumes": {
  "Count": 3,
  "DataPoints": 108,
  "PollTime": 432852,
  "ApiTime": 401694,
  "ParseTime": 15296,
  "PluginTime": 305
 },
 "Svms": {
  "Count": 0,
  "DataPoints": 0,
  "PollTime": 0,
  "ApiTime": 0,
  "ParseTime": 0,
  "PluginTime": 0
 },
 "Platform": {
  "OS": "",
  "Arch": "",
  "MemoryKb": 0,
  "CPUs": 0
 },
 "Harvest": {
  "UUID": "ab87c39d-bf08-4b72-a74d-6418514df889",
  "Version": "21.08.1318",
  "Release": "v21.05.4",
  "Commit": "ab48bed",
  "BuildDate": "2021-08-13T18:39:54+0530",
  "NumClusters": 1
 }
}
}
}

cmd/tools/asup/asup.go
Hardikl: changed all exec command to use library

cgrinds:
- [ ] Ensue changes in `cmd/collectors/zapi/collector/zapi.go` work with 7-mode

cmd/poller/poller.go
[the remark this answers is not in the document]
cgrinds: is it time to change this to once a month

cmd/poller/poller.go
```diff
@@ -281,13 +285,69 @@ func (p *Poller) Init() error {
...
+
 	// famous last words
 	logger.Info().Msg("poller start-up complete")
 
 	return nil
 
 }
 
+// to the asup
+func (p *Poller) startAsup() (*matrix.Matrix, error) {
```
cgrinds, line 342: is the return correct? looks like we only ever return `nil` for the Matrix - should this method be `func (p *Poller) startAsup() error {` instead?

pkg/conf/conf.go
cgrinds: is this still needed with sha1 of cluster-uuid?

pkg/conf/conf.go
```diff
@@ -396,6 +425,8 @@ type Consul struct {
 
 type Tools struct {
 	GrafanaApiToken *string `yaml:"grafana_api_token,omitempty"`
+	AsupEnabled     *bool   `yaml:"asup_messaging_enabled,omitempty"`
+	HarvestUUID     *string `yaml:"harvest_uuid,omitempty"`
```
cgrinds, line 429: is this still needed with sha1 of cluster-uuid

cmd/poller/poller.go
Hardikl: This function is invoked by scheduler and it's signature should be func name() (*matrix.Matrix, error). That's the reason for this return.

Hardikl:
"Target": {
  "Version": "9.10.1",
  "Model": "cdot",
  "Serial": "1-80-000011",
  "Ping": 0,
  "ClusterUuid": "3ccba2b0-f8f7-11eb-be9f-005056a70771"
 },


"Harvest": {
  "UUID": "1e6d0f0f858cb84c723a93c65cfe472215c326a5",
  "Version": "21.08.1620",
  "Release": "v21.05.4",
  "Commit": "a09f3f0",
  "BuildDate": "2021-08-16T20:51:20+0530",
  "NumClusters": 1
 }

cgrinds (commented):
Looks good @Hardikl - there are a few changes need. I'll push a commit and we can discuss

cmd/poller/poller.go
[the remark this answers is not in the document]
Hardikl: Yes, done.
Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h". So changed to 720h which is 30days period

pkg/conf/conf.go
Hardikl: not needed, removed.

pkg/conf/conf.go
Hardikl: this is removed as not needed.

Hardikl:
> * [ ]  Ensue changes in `cmd/collectors/zapi/collector/zapi.go` work with 7-mode

Yes, 7 mode would be handled here: https://github.com/NetApp/harvest/blob/main/pkg/api/ontapi/zapi/system.go#L23

cmd/tools/asup/asup.go
```diff
@@ -85,76 +84,91 @@ func DoAsupMessage(config string, collectors []collector.Collector, status *matr
...
+	}
+	return nil
+}
+
+func sendAsupVia(msg *asupMessage, pollerName string, asupExecPath string) error {
 	var (
 		payloadPath string
 		err         error
 	)
 	asupTimeOutLimit := 10 * time.Second
```
Hardikl, line 115: asup exec timeout is also need to decide may be 5 mins ?

pkg/conf/conf.go
```diff
@@ -407,7 +407,7 @@ type Consul struct {
 
 type Tools struct {
 	GrafanaApiToken *string `yaml:"grafana_api_token,omitempty"`
-	AsupEnabled     *bool   `yaml:"asup_messaging_enabled,omitempty"`
+	AsupDisabled    bool    `yaml:"harvest_no_asup,omitempty"`
```
Hardikl, line 410: default would be always enabled?

cmd/poller/poller.go
```diff
@@ -285,26 +283,27 @@ func (p *Poller) Init() error {
...
+	} else {
 		if !p.targetIsOntap() {
-			logger.Info().Msg("disabling ASUP messaging as target is not ONTAP")
-		} else if err = p.schedule.NewTaskString("asup", asupInterval, p.startAsup); err != nil {
-			logger.Error().Stack().Err(err).Msg("set ASUP interval:")
+			logger.Info().
+				Str("poller", p.name).
+				Msg("Autosupport disabled since poller not connected to ONTAP.")
+		}
+		if err = p.schedule.NewTaskString("asup", asupInterval, p.startAsup); err != nil {
```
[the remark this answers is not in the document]
cgrinds: good catch

cmd/tools/asup/asup.go
cgrinds: how does exec timeout differ from `asupTimeOutLimit`

cmd/tools/asup/asup.go
```diff
@@ -0,0 +1,441 @@
...
+	if err = connection.Init(2); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	return connection
+}
+
+func getSha1Uuid(clusterUuid string) string {
+	shaHash := sha1.New()
```
rahulguptajss, line 435: why are we even doing sha1 on this. This value is anyways included in payload.

pkg/conf/conf.go
rahulguptajss: naming convention looks other way around.

cmd/poller/poller.go
```diff
@@ -281,13 +283,48 @@ func (p *Poller) Init() error {
...
+func (p *Poller) startAsup() (*matrix.Matrix, error) {
+	if p.collectors != nil {
+		if err := asup.DoAsupMessage(p.options.Config, p.collectors, p.status, p.name); err != nil {
+			logger.Error().Err(err).
+				Str("poller", p.name).
+				Msg("Start autosupport failed.")
+			return nil, err
+		}
+	}
+	return nil, nil
```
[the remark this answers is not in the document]
cgrinds: required by the interface method signature - i had the same question in an earlier review

cmd/tools/asup/asup.go
cgrinds: if we don't cluster-uuid and harvest-uuid will be in conflict

cmd/tools/asup/asup.go
Hardikl: actually this is the exec timeout we have provided in context at line 135

cmd/poller/poller.go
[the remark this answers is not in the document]
Hardikl: yups it's expected for scheduled method

rahulguptajss (approved):

rahulguptajss (approved):

PR #398, opened 2021-08-08
refactor: remove PrometheusConsul from test

cgrinds:
Re-add when PrometheusConsul is implemented. Until then, it's confusing to include as an example

### DIFF
--- a/cmd/tools/doctor/doctor.go
+++ b/cmd/tools/doctor/doctor.go
@@ -87,7 +87,7 @@ func checkExporterTypes(config conf.HarvestConfig) validation {
 			continue
 		}
 		switch *exporter.Type {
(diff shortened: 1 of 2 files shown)


Hardikl (dismissed):
do we need to remove from here : https://github.com/NetApp/harvest/blob/main/cmd/tools/doctor/doctor.go#L90 ?

vgratian (dismissed):

cdurai-netapp (approved):

cgrinds:
> do we need to remove from here : https://github.com/NetApp/harvest/blob/main/cmd/tools/doctor/doctor.go#L90 ?

good catch - yes, it should be removed from there too

PR #399, opened 2021-08-08
fix: built-in collectors should have overridable config

### LINKED ISSUE #396: Failed to define custom ZapiPerf collector file (other than defaults.yaml)
**Describe the bug**
When I try to change the collector configuration file for ZapiPerf like describe at the end of the section (change to not use the default.yaml) : https://github.com/NetApp/harvest/blob/main/cmd/collectors/zapiperf/README.md#harvest-configuration-file
I got the error : "cannot unmarshal !!map into string"

**Environment**

- Harvest version: harvest version 21.05.4-1 (commit 8976370) (build date 2021-08-02T11:24:55+0200) linux/amd64
- Command line arguments used: ./bin/harvest start --foreground --verbose--collectors ZapiPerf --objects SystemNode DC-01
- OS: Debian 10.9
- Install method: dpkg install from manual package compilation 
- ONTAP Version: CDOT 9.5

**To Reproduce**

Create a file limited.yaml in folder /opt/harvest/conf/zapiperf/ : 
```
root@XXX:/opt/harvest# ll conf/zapiperf/
total 16
drwxr-xr-x 3 harvest harvest 4096 juil.  1 22:33 7mode
drwxr-xr-x 4 harvest harvest 4096 août   6 16:59 cdot
-rw-r--r-- 1 harvest harvest 1798 août   2 11:24 default.yaml
-rw-r--r-- 1 harvest harvest   95 août   6 15:44 limited.yaml
```

With the following content: 
```
root@XXX:/opt/harvest# cat conf/zapiperf/limited.yaml
objects:
  SystemNode:             system_node.yaml
  HostAdapter:            hostadapter.yaml
```

And change the harvest.yml on the application root folder : 
```
root@XXX:/opt/harvest# cat harvest.yml
...

Defaults:
  collectors:
    - ZapiPerf
  use_insecure_tls: true

Pollers:
  DC-01:
    datacenter: Lab
    addr: 192.168.xxx.xxx
    auth_style: basic_auth
    username: admin
    password: xxxxxxx
    exporters:
      - my_influxdb
    collectors:
      - ZapiPerf:
        - limited.yaml
```

Got the error when a try to launch with command : 
```
root@XXX:/opt/harvest# ./bin/harvest start --foreground --trace --collectors ZapiPerf --objects SystemNode DC-01
error unmarshalling config file=[./harvest.yml] yaml: unmarshal errors:
  line 25: cannot unmarshal !!map into string
config [./harvest.yml]: yaml: unmarshal errors:
  line 25: cannot unmarshal !!map into string

```

Line 25 corresponds to:
```
      - ZapiPerf:
```

cgrinds:
Fixes #396

### DIFF
--- a/pkg/conf/conf.go
+++ b/pkg/conf/conf.go
@@ -398,23 +398,28 @@ type Tools struct {
 	GrafanaApiToken *string `yaml:"grafana_api_token,omitempty"`
 }
 
+type Collector struct {
+	Name      string
+	Templates *[]string
+}
+
 type Poller struct {
-	Datacenter     *string   `yaml:"datacenter,omitempty"`
-	Addr           *string   `yaml:"addr,omitempty"`
-	AuthStyle      *string   `yaml:"auth_style,omitempty"`
-	Username       *string   `yaml:"username,omitempty"`
-	Password       string    `yaml:"password,omitempty"`
-	UseInsecureTls *bool     `yaml:"use_insecure_tls,omitempty"`
-	SslCert        *string   `yaml:"ssl_cert,omitempty"`
-	SslKey         *string   `yaml:"ssl_key,omitempty"`
-	LogMaxBytes    *int64    `yaml:"log_max_bytes,omitempty"`
-	LogMaxFiles    *int      `yaml:"log_max_files,omitempty"`
-	Exporters      *[]string `yaml:"exporters,omitempty"`
-	Collectors     *[]string `yaml:"collectors,omitempty"`
-	IsKfs          *bool     `yaml:"is_kfs,omitempty"`
-	PollerSchedule *string   `yaml:"poller_schedule,omitempty"`
-	ClientTimeout  *string   `yaml:"client_timeout,omitempty"`
-	LogSet         *[]string `yaml:"log,omitempty"`
+	Datacenter     *string      `yaml:"datacenter,omitempty"`
+	Addr           *string      `yaml:"addr,omitempty"`
+	AuthStyle      *string      `yaml:"auth_style,omitempty"`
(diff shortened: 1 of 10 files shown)


Hardikl (approved):

pkg/conf/conf.go
```diff
@@ -448,6 +453,24 @@ type Pollers struct {
 	namesInOrder []string
 }
 
+var defaultTemplate = &[]string{"default.yaml"}
```
vgratian, line 456: let's make this line into:
`var defaultTemplate &[]string{"default.yaml", "custom.yaml"}`
then delete lines  495-497 from poller.go

vgratian:
Here is my two cents to document this changes. E.g. in `README.md`, I would change the section [Configuring Collectors](https://github.com/NetApp/harvest#configuring-collectors), to something like this:

> Collectors are configured by their own configuration files (*templates*), which are stored in subdirectories in [conf/](conf/). Most Harvest collectors are "object-oriented", which means multiple instances of the same collector will concurrently collect a subset of metrics that are related to a logical subsystem or resource. In such case, templates will contain a list of objects that are in turn defined in "subtemplates". 
>
> At start-up, Harvest will look for a file, `default.yaml`, in the conf-subdirectory of the collector. Additionally, if a user-created `custom.yaml` is found, it will be merged to the former. This default behavior can be overridden in `config.yml`, for an example see [here](pkg/conf/testdata/issue_396.yaml).
>
> For a list of collector-specific parameters, see the documentation of each collector.

Btw, thanks a lot for the PR, this fix was long due.

cgrinds:
Updated `README.md` and addressed review comments

PR #412, opened 2021-08-13
fix: disk dashboard status column

Hardikl:
local testing grafana: http://10.249.28.157:3000/.  [used grafana version for testing: 7.5.10]

Covered these:
1. status column issue
2. disk utilization percentage

I don't think we have issue with 'Disk and Take Drives Throughput' as metric values are in µs and it should be.


cgrinds (approved):
thanks @Hardikl

PR #413, opened 2021-08-13
feat: templates should be extendable without modifying the original

### LINKED ISSUE #394: Templates should be extendible without modifying the original
Today it's possible to add new templates without modifying the original, but it's not possible to add extra properties to one of the existing templates.

At the moment, values from `default.yaml` will override key/values in `custom.yaml` which means if your `custom.yaml` contains something like below, it will be ignored. We should improve extending - for the moment, the workaround is modify `aggr.yaml`.

```
objects:
  Aggregate: aggr-plus.yaml
```

See comments [here](https://github.com/NetApp/harvest/issues/384#issuecomment-893874327)

rahulguptajss:
fixes #394

### DIFF
--- a/pkg/test/node_test.go
+++ b/pkg/test/node_test.go
@@ -1,0 +1,47 @@
+package test
+
+import (
+	"goharvest2/pkg/tree"
+	"testing"
+)
+
+func TestNode_Merge(t *testing.T) {
+	defaultTemplate, _ := tree.Import("yaml", "testdata/default.yaml")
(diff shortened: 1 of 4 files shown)


rahulguptajss:
@cgrinds Could you pls point me to the doc which needs to be updated. Doc @ https://github.com/NetApp/harvest/blob/main/cmd/collectors/zapi/README.md#example-subtemplate looks ok

cgrinds (approved):
one line code change and then 74 lines of test code. nice!

PR #416, opened 2021-08-16
new SVM panels for NFSv4 and NFSv4.1

burkl:
Grafana panels for visualizing NFSv4x counters from Harvest 2

### DIFF
--- a/grafana/dashboards/harvest_dashboard_svm.json
+++ b/grafana/dashboards/harvest_dashboard_svm.json
@@ -59,11 +59,11 @@
   "gnetId": null,
   "graphTooltip": 0,
   "id": null,
-  "iteration": 1626948102350,
+  "iteration": 1628866003817,
   "links": [],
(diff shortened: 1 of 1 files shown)


cgrinds:
thanks for the contribution @burkl !

this looks great, will will check and make sure it works on earlier versions of Grafana since Harvest supports `7.4.2` or higher

burkl:
@cla-bot check

cgrinds:
Should there be an All dropdown in the SVM filter and should that be the default when opening the dashboard similar to, e.g. Aggregate?

image

burkl:
@cgrinds not sure, if there is a counter for NFS stats for all SVMs.

Actually, I used the Harvest 1.6 Grafana Dashboard as a template which had no "All" option

cgrinds:
thanks @burkl we'll check if "All" is needed as part of the review

PR #429, opened 2021-08-19
feat: asup binary changes part 1

rahulguptajss:


rahulguptajss:
make asup package GIT_TOKEN=<<token>>

Makefile
```diff
@@ -119,3 +124,13 @@ dist-tar:
...
 	@echo "tar artifact @" ${DIST}/${HARVEST_PACKAGE}.tar.gz
+
+asup: ## Build harvest asup binary
+	@echo "Building Asup"
+	@rm -rf bin/asup
+	@rm -rf ${ASUP_TMP}
+	@mkdir ${ASUP_TMP}
+	@git clone https://${GIT_TOKEN}@github.com/NetApp/harvest-private.git ${ASUP_TMP}
+	@cd ${ASUP_TMP}/harvest-asup && make ${ASUP_MAKE_TARGET} VERSION=$VERSION RELEASE=$RELEASE
+	@mkdir -p ${CURRENT_DIR}/bin
```
Hardikl, line 135: mkdir may not needed, as we would use same location where other binary resides.
hardikl@hardikl-mac-0 harvest % ls -ltr bin 
total 132544
-rwxr-xr-x  1 hardikl  staff  12637504 Aug 19 15:10 harvest
-rwxr-xr-x  1 hardikl  staff   9580048 Aug 19 15:10 asup
-rwxr-xr-x  1 hardikl  staff  14068064 Aug 19 15:10 poller
-rwxr-xr-x  1 hardikl  staff     13448 Aug 19 15:10 daemonize
-rwxr-xr-x  1 hardikl  staff  10434032 Aug 19 15:10 zapi
-rwxr-xr-x  1 hardikl  staff   9792544 Aug 19 15:10 grafana
-rwxr-xr-x  1 hardikl  staff  11322656 Aug 19 15:10 rest
hardikl@hardikl-mac-0 harvest %

rahulguptajss: yes but consider a scenario where you just may want to do make asup. It has -p flag in case already exists.

Hardikl: Agreed. 

Also may be we need to add this task in build?

rahulguptajss: nope. I don't expect open source community to build asup binary as it will anyways fail for them. Hence kept it as seperate target. Open to ideas if we write make file in such a way where asup target is allowed to fail inside build target.

cgrinds (approved):

PR #435, opened 2021-08-20
fix: reset metadata for each task

vgratian:
fix makes sure that metadata of each task is reset only when it's being executed.

other minor changes:
- remove dead metadata metrics: `content_length` and `api_time_percent`
- set the `poll_time` and `task_time` metrics for all tasks, not just `data`


cgrinds (approved):
looks good and better than the alternatives we discussed

PR #436, opened 2021-08-20
fix: reset collector retryDelay when connection resumes

### LINKED ISSUE #356: Behaviour with zapi collector data schedule - After losing cluster comms
**Describe the bug**
The default data schedule for the zapi collector is 180s defined at `conf/zapi/default.yaml`. By default when using the Prometheus exporter the data is always available after the collector starts regardless of the schedule, if these collectors fail and then recover (in my case, we lost comms to one of the clusters), the exporter begins to only report data samples for Zapi metrics every 180s, when it should be returning cached instances between those intervals.

**Environment**

- Harvest version: harvest version 21.05.4-2 (commit 19f8f25) (build date 2021-07-22T15:41:23+0000) linux/amd64
- OS: RHEL 7.9
- Install method: yum
- ONTAP Version: 9.4P8
- Other:

**To Reproduce**
Lose connection to a cluster (possibly for longer than the schedule interval? at least thats what happened for us). 

**Expected behavior**
Zapi Collectors should recover and cache data appropriately for the prometheus exporter

**Actual behavior**
Zapi Collectors do not cache data and the exporter only returns samples according to the schedule interval.

Restarting harvest resumes functions correctly.

**Additional context**
There's nothing in the logs to indicate any issues, funnily enough when harvest loses connection to a cluster, the only logging thats shown is context deadline exceeded for `ZapiPerf` collectors and zero logging for `Zapi` collectors.

**EDIT**
This seems to affect zapi perf as well, I'm actually not sure about the correlation between the data schedules, all I can say is that after collector recovery they seem to only produce metrics every 15mins.

cgrinds:
- add more logging when connection fails

Fixes #356

### DIFF
--- a/cmd/poller/collector/collector.go
+++ b/cmd/poller/collector/collector.go
@@ -312,6 +312,10 @@ func (me *AbstractCollector) Start(wg *sync.WaitGroup) {
 						//logger.Error(me.Prefix, err.Error())
 						me.Logger.Warn().Msgf("target unreachable, entering standby mode (retry to connect in %d s)", retryDelay)
 					}
+					me.Logger.Debug().
+						Int("retryDelay", retryDelay).
+						Err(err).
+						Msg("Target unreachable, entering standby mode (retry in retryDelay s)")
 					me.Schedule.SetStandByMode(task, time.Duration(retryDelay)*time.Second)
 					me.SetStatus(1, errors.ERR_CONNECTION)
 				// there are no instances to collect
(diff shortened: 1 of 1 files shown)


cmd/poller/collector/collector.go
```diff
@@ -339,6 +343,7 @@ func (me *AbstractCollector) Start(wg *sync.WaitGroup) {
 			} else if me.Schedule.IsStandBy() {
 				// recover from standby mode
 				me.Schedule.Recover()
+				retryDelay = 1
```
vgratian, line 346: good catch!
I had completely overseen this..

vgratian (approved):

cmd/poller/collector/collector.go
cgrinds: all credit to hashi825 for noticing and pointing out - i did the easy part 😄

Issue #473, opened 2021-08-27
Counters from zapiperf object_store_client_op.yaml not collected


jgasher:
**Describe the bug**
Counters from object_store_client_op.yaml are not being collected.

**Environment**
Provide accurate information about the environment to help us reproduce the issue.

- Harvest version:  harvest version 21.08.2409-v21.05.4 (commit e8f580d) (build date 2021-08-24T09:46:04-0400) linux/amd64

- Command line arguments used: bin/harvest start
- OS: CENTOS 8
- Install method: native
- ONTAP Version: 9.8+
- Other:

**To Reproduce**
Use Harvest

**Expected behavior**
Tell us what should happen:
This should be collected:
counters:
  - instance_name
  - instance_uuid
  - node_name              => node
  - average_latency
  - get_throughput_bytes
  - put_throughput_bytes
  - stats
  - throughput_ops

**Actual behavior**
Tell us what happens instead:
These aren't collected.

**Possible solution, workaround, fix**
Is it possible these aren't collected because the fabricpool.yaml and object_store_client_op.yaml both output as the fabricpool object?

**Additional context**
I have several systems internal to NetApp labs that can be used to demonstrate the behavior.

rahulguptajss:
@jgasher Is fabricpool.yaml a new template added from your side? I do not see it in default template list. Though having same object name in multiple templates causes issue as described by you. #462 also highlights the same problem. We'll be fixing it in upcoming release.

cgrinds:
Verified on `4044fcf` 

```
curl -s 'http://localhost:12993/metrics' | grep fabricpool | grep -v cloud
fabricpool_throughput_ops{datacenter="4219-ATT",cluster="primary",node="a400-02",fabricpool="0"} 0.6501261846760531
fabricpool_throughput_ops{datacenter="4219-ATT",cluster="primary",node="a400-01",fabricpool="0"} 0.05000970651354255
```

PR #477, opened 2021-08-30
fix: metrics overwrite problem

### LINKED ISSUE #462: NFS metrics appear to overwrite themselves
**Describe the bug**
There is a strange behaviour on clusters containing both nfsv3 and nfsv4.1 (or 4 I spose), svm and node related NFS metrics that use `global_labels` nfsv appear to overwrite each other/not appear.

**Environment**

- Harvest version: harvest version 21.05.4-2 (commit 19f8f25) (build date 2021-07-22T15:41:23+0000) linux/amd64
- OS: RHEL 7.9
- Install method: yum
- ONTAP Version: 9.4P8
- Other:

**To Reproduce**
Enable both nfsv3.yaml and nfsv4_1 or nfsv4.yaml in the default template.

**Expected behavior**
Metrics should export metrics relating to any version of nfs

**Actual behavior**
Metrics collected from the Prometheus exporter from a cluster having SVM/Node nfs with differing versions appear to alternate after a while which metric gets exported.
ie:

`node_nfs_access_avg_latency{cluster="cluster1",node="node1",nfsv="v3"} 221`

sometime later, the above goes missing and you get the v4 (or v4.1) related metric

`node_nfs_access_avg_latency{cluster="cluster1",node="node1",nfsv="v4.1"} 111`

Unsure if its do with the Prometheus exporter itself or the code around the use of `global_labels`, disabling one or the other nfs template in the default.yaml fixes the issue for that particular nfs version, this requires disabling both node level and svm level metrics. If only one is disabled, once node nfsv4 metrics get exported, then svm nfsv3 metrics disappear.

**Additional context**
N/A

rahulguptajss:
Fixes #462 #473

### DIFF
--- a/cmd/collectors/zapi/plugins/shelf/shelf.go
+++ b/cmd/collectors/zapi/plugins/shelf/shelf.go
@@ -73,7 +73,7 @@ func (my *Shelf) Init() error {
 
 		my.instanceLabels[attribute] = dict.New()
 
-		my.data[attribute] = matrix.New(my.Parent+".Shelf", "shelf_"+objectName)
+		my.data[attribute] = matrix.New(my.Parent+".Shelf", "shelf_"+objectName, "shelf_"+objectName)
 		my.data[attribute].SetGlobalLabel("datacenter", my.ParentParams.GetChildContentS("datacenter"))
 		my.data[attribute].SetGlobalLabel("cluster", my.client.Name())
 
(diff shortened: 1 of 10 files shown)


vgratian (approved):
Good job for finding and fixing the issue.

I think at some point we need to revisit how the matrix is initialized/utilized. Its interface can be simpler and safer than it is now. We want to avoid that in the future developers/contributors get confused by these arguments and run into the same confusion (mea culpa).

The challenge is that these identifiers of the matrix should not only work for Harvest itself, but also for the databases where the exporters push the data.

cgrinds (approved):

PR #482, opened 2021-08-30
feat: svm details dashboard

cgrinds:
Thanks to Jeff Asher for sharing!

### DIFF
--- a/grafana/dashboards/harvest_dashboard_svm_details.json
+++ b/grafana/dashboards/harvest_dashboard_svm_details.json
@@ -14,7 +14,7 @@
       "type": "grafana",
       "id": "grafana",
       "name": "Grafana",
-      "version": "8.0.6"
+      "version": "8.1.1"
     },
     {
       "type": "panel",
@@ -50,6 +50,12 @@
         "hide": true,
         "iconColor": "rgba(0, 211, 255, 1)",
         "name": "Annotations & Alerts",
+        "target": {
+          "limit": 100,
+          "matchAny": false,
+          "tags": [],
+          "type": "dashboard"
+        },
         "type": "dashboard"
       }
     ]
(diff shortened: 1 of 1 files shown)


Hardikl (dismissed):

vgratian (changes requested):
Great work, I noticed a few, mostly cosmetic issues that need to be fixed:

## Variables
The SVM variable uses the metric `volume_labels`. This does not give us *all* SVMs in the cluster. I could not find a solution for this. (We could just drop the metric, but then I don't know how to filter on cluster).

## Most Panels
- Since the SVM variable is single-select only, it is slightly incorrect to have `svm=~"$SVM"` in the queries. This might match similarly-named SVMs and makes the queries slightly more expensive.
- Latency gauges and graphs divide metric value by `1000`. This is redundant: makes the query results slightly lossy and expensive. Instead we should just use `microseconds` as unit.

## CIFS Frontend Drilldown 
- The Read Latency gauge is duplicate and has query for average latency. Should probably query throughput.

## NFSv3 Frontend Drilldown 
- Throughput (Data) gauges and graphs: unit should be `bytes/sec` instead of `kilobytes/sec`
    - Read and Write Latency graph is duplicate, probably needs to be Throughput.

## NFSv4 Frontend Drilldown 
- Throughput (Data) gauges need to have `bytes/sec` as unit

## NFSv4.1 Frontend Drilldown 
- Throughput (Data) gauges need to have `bytes/sec` as unit

## QoS Policy Group Drilldown
- (minor) Latency gauges have only base threshold

vgratian:
I'll push a commit with fixes.

vgratian (approved):

cgrinds:
these look great. Thanks Jeff and @vgratian

PR #488, opened 2021-09-01
feat: rpm upgrade test with other changes

cdurai-netapp:
1. add upgrade workflow in Jenkins pipeline
2. upgrade test validation for rpm Package

### DIFF
--- a/jenkins/artifacts/jenkinsfile
+++ b/jenkins/artifacts/jenkinsfile
@@ -55,7 +55,7 @@ pipeline {
         stage('Setup GO') {
             steps {
                 sh '''
-                wget -O go.tar.gz "https://dl.google.com/go/$(curl https://golang.org/VERSION?m=text).linux-amd64.tar.gz"
+                wget -q -O go.tar.gz "https://dl.google.com/go/$(curl https://golang.org/VERSION?m=text).linux-amd64.tar.gz"
                 rm -rf /usr/local/go && tar -C /usr/local -xzf go.tar.gz
                 '''
             }
@@ -170,9 +170,8 @@ pipeline {
     }
 }
 
-
 def getBranchName(gitBranchName, paramBranchName) {
-    if (gitBranchName?.trim()) {
+    if (gitBranchName?.trim() && gitBranchName != "origin/main") {
         return gitBranchName
     }
     return paramBranchName


cgrinds:
Thanks @cdurai-netapp looks good. I don't think the upgrade test does this yet, but one area we've seen upgrade fail is when pollers are already running, and then an upgrade is started. Can you add a test for that? I don't recall if the upgrade stops existing, leaves them running, etc.

cgrinds (approved):

integration/Jenkinsfile
```diff
@@ -111,25 +106,28 @@ pipeline {
                                 '''
                             }
                         }
-
                         stage('Install RPM pkg') {
                             steps {
                                 sh '''
                                 echo "Installing"
-                                export PATH=$PATH:/usr/local/go/bin
+                                export BUILD_PATH=https://github.com/NetApp/harvest/releases/download/v21.08.0/harvest-21.08.0-6.x86_64.rpm
```
rahulguptajss, line 113: downloading from github will cause our metrics to skewed @ https://hanadigital.github.io/grev/?user=NetApp&repo=harvest

May be copy it locally?

cdurai-netapp: good catch. fixed.

jenkins/artifacts/JenkinWorkflow
```diff
@@ -1,112 +1,137 @@
 pipeline {
     agent {label 'buildserver'}
 
+    parameters {
```
rahulguptajss, line 4: can we combine these 2 jenkins file? most lines are same?

cgrinds: @cdurai-netapp was this commented addressed? do you think this makes sense?

cdurai-netapp: yes it make sense. working on it.

cdurai-netapp: fixed and sent it for review.

jenkins/artifacts/jenkinsfile
```diff
@@ -22,9 +23,7 @@ pipeline {
         RELEASE =sh (returnStdout: true, script: """
         echo \"${params.RELEASE}\"
         """).trim()
-        BRANCH =sh (returnStdout: true, script: """
-        echo \"${params.BRANCH}\"
-        """).trim()
+        BRANCH = getBranchName(env.GIT_BRANCH, params.BRANCH)
```
rahulguptajss, line 26: Why do we need to check between env and params? Why don't we keep it just read from params?

cdurai-netapp: since we use same jenkins file, we have to add a logic to get dev branch when we run this jenkins file in a multibranch pipeline jobs.  GIT_BRANCH is an inbuilt env variable which will return dev branch info. hence i have added this

cgrinds (approved):

rahulguptajss (approved):

PR #489, opened 2021-09-02
feat: split dashboards for 7mode - part1

Hardikl:
Added global label only in 7mode node.yaml, we would use the node_labels{system_type} to take decision of cmode/7mode.

Identified gaps:
- 7 mode don't have product-version in system-get-info zapi
- 7 mode don't have tx_hw_errors, rx_alignment_errors, rx_crc_errors, rx_length_errors in nic_common object

Changes done:
Cluster
- node_new_status to cluster_new_status

Aggregate
- Physical Space Used to space used 
- removed Capacity Tier Used(aggr_space_capacity_tier_used), Data Compaction space used(aggr_space_data_compaction_saved), Space Reserved for Snapshot %(space reserved for snapshot) panels which are nor available in 7 mode zapi response. 

Network
- Removed Transmission interrupts, Transmission errors, Link Down, Link Failure panels 

Network with Nvme/FC
- Removed Nvme/FC receive, Nvme/FC send, Transmission interrupts, Transmission errors, Link Down, Link Failure panels, Nvme/Fc ports table

### DIFF
--- a/conf/zapi/7mode/8.6.0/node.yaml
+++ b/conf/zapi/7mode/8.6.0/node.yaml
@@ -2,6 +2,9 @@ name:                       Node
 query:                      system-get-info
 object:                     node
 
+global_labels:
+  - system_type: 7mode
+
 counters:
   system-info:
     - memory-size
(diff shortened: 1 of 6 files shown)


Hardikl:
cmode dashboard would show only cmode datacenter
image

7mode dashboard would show only 7mode datacenter
image

image

cgrinds:
Looks good @Hardikl - as discussed in chat, dropping here so we don't forget. If customers have customized dashboards and they don't update the queries in those dashboards (as done in this PR), then they will still see 7-mode clusters in their custom dashboards

cgrinds (approved):

PR #514, opened 2021-09-23
fix: escape whitespace in InfluxDB exporter

cdurai-netapp:
Add url encoding for bucket and org parameter in InfluxDB parameters

### DIFF
--- a/cmd/exporters/influxdb/influxdb.go
+++ b/cmd/exporters/influxdb/influxdb.go
@@ -10,6 +10,7 @@ import (
 	"goharvest2/pkg/color"
 	"goharvest2/pkg/errors"
 	"goharvest2/pkg/matrix"
+	"goharvest2/pkg/util"
 	"io/ioutil"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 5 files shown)


Hardikl (approved):

cmd/exporters/influxdb/influxdb.go
```diff
@@ -112,7 +113,8 @@ func (e *InfluxDB) Init() error {
 
 		urlToUSe := "http://" + *addr + ":" + strconv.Itoa(*port)
 		url = &urlToUSe
-		e.url = fmt.Sprintf("%s/api/v%s/write?org=%s&bucket=%s&precision=%s", *url, *version, *org, *bucket, *precision)
+		e.url = fmt.Sprintf("%s/api/v%s/write?org=%s&bucket=%s&precision=%s", *url, *version, util.EscapeUrl(*org),
```
cgrinds, line 116: does the EscapeUrl function carry enough weight to add or can we call url.PathEscape directly?

cgrinds (commented):
thanks @cdurai-netapp can you add a unit test that covers this. Please add a test case that verifies this does what we want and make sure the PR mentions the issue so a connection is made between the PR and issue

cmd/exporters/influxdb/influxdb.go
cdurai-netapp: i cant directly use url.PathEscape since url is already consumed by  influxdb.go as a variable. hence i coudlnt directly use it . otherwise i have to rename e.url everywhere seems like

cdurai-netapp: sure. will add unit test for this.

PR #530, opened 2021-09-28
fix: poller should parse busybox ping

### LINKED ISSUE #529: Poller does not parse the contents of `ping` correctly when it use ping from busybox.
**Describe the bug**
Poller does not parse the contents of `ping` correctly when it use ping from busybox. That causes metric `metadata_target_status` report incorrectly. 
The output of ping doesn't includes `mdev =` and that causes ping() returns false. 

**Environment**

- Harvest version: harvest version 2.0-nightly (commit ) (build date 2021-07-22T15:42:55+0000) linux/amd64
- Command line arguments used: docker image from https://hub.docker.com/r/rahulguptajss/harvest

**To Reproduce**

```
bash-5.1# ps -ef |grep poller
31257 root      2d18 bin/poller --poller netapp --promPort 12990

bash-5.1# cat harvest.yml |grep netapp -A 2
  netapp:
    datacenter: test
    addr: 192.168.1.133

bash-5.1# wget 127.0.0.1:12990/metrics
bash-5.1# cat metrics |grep metadata_target_status
metadata_target_status{version="2.0",hostname="netapp-metrics-harvest-68f4db67d8-958c7",poller="netapp",addr="192.168.1.133"} 1
```

There is no `mdev` in outupt of ping from busybox. 

```
bash-5.1# ping 192.168.1.133 -w 5 -c 1 -q
PING 192.168.1..133 (192.168.1.133): 56 data bytes

--- 192.168.1.133 ping statistics ---
1 packets transmitted, 1 packets received, 0% packet loss
round-trip min/avg/max = 0.291/0.291/0.291 ms

bash-5.1# ping --help
BusyBox v1.33.1 () multi-call binary.

Usage: ping [OPTIONS] HOST

Send ICMP ECHO_REQUESTs to HOST

        -4,-6           Force IP or IPv6 name resolution
        -c CNT          Send only CNT pings
        -s SIZE         Send SIZE data bytes in packets (default 56)
        -i SECS         Interval
        -A              Ping as soon as reply is recevied
        -t TTL          Set TTL
        -I IFACE/IP     Source interface or IP address
        -W SEC          Seconds to wait for the first response (default 10)
                        (after all -c CNT packets are sent)
        -w SEC          Seconds until ping exits (default:infinite)
                        (can exit earlier with -c CNT)
        -q              Quiet, only display output at start/finish
        -p HEXBYTE      Payload pattern
bash-5.1# 
```

If you run the ping command from a normal host (not using busybox). You'll abel to see `mdev`.

```
$ ping 192.168.1.133 -w 5 -c 1 -q
PING 192.168.1.133 (192.168.1.133) 56(84) bytes of data.

--- 10.181.193.133 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 0.155/0.155/0.155/0.000 ms

$ ping -V
ping utility, iputils-s20160308
```

**Expected behavior**

```
bash-5.1# wget 127.0.0.1:12990/metrics
bash-5.1# cat metrics |grep metadata_target_status
metadata_target_status{version="2.0",hostname="netapp-metrics-harvest-68f4db67d8-958c7",poller="netapp",addr="192.168.1.133"} 0
```

The paring of contents needs to improve to prevent this issue. There is no `mdev =` in the output of ping when it is from busybox. 

https://github.com/NetApp/harvest/blob/214f702b11eac30fcfdc351d4173e691643fc2d5/cmd/poller/poller.go#L506-L516

cgrinds:
Fixes #529

### DIFF
--- a/cmd/poller/poller.go
+++ b/cmd/poller/poller.go
@@ -87,7 +87,7 @@ var deprecatedCollectors = map[string]string{
 	"psutil": "Unix",
 }
 
-var pingRegex = regexp.MustCompile(`round-trip min/avg/max = (.*?)/`)
+var pingRegex = regexp.MustCompile(` = (.*?)/`)
 
 // Poller is the instance that starts and monitors a
 // group of collectors and exporters as a single UNIX process
@@ -515,19 +515,12 @@ func (p *Poller) ping() (float32, bool) {
 }
 
 func (p *Poller) parsePing(out string) (float32, bool) {
-	// Non BusyBox systems include mdev in output
-	if x := strings.Split(out, "mdev = "); len(x) > 1 {
-		if y := strings.Split(x[len(x)-1], "/"); len(y) > 1 {
-			if p, err := strconv.ParseFloat(y[0], 32); err == nil {
+	if strings.Contains(out, "min/avg/max") {
+		match := pingRegex.FindStringSubmatch(out)
+		if len(match) > 0 {
+			if p, err := strconv.ParseFloat(match[1], 32); err == nil {
 				return float32(p), true
 			}
-		}
-	}
-	// BusyBox does not include mdev
-	match := pingRegex.FindStringSubmatch(out)
-	if len(match) > 0 {
-		if p, err := strconv.ParseFloat(match[1], 32); err == nil {
-			return float32(p), true
 		}
 	}
 	return 0, false
--- a/cmd/poller/poller.go
+++ b/cmd/poller/poller.go
@@ -50,6 +50,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path"
+	"regexp"
 	"runtime"
 	"runtime/debug"
 	"strconv"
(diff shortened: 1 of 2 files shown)


cdurai-netapp (dismissed):

cmd/poller/poller_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+func TestPingParsing(t *testing.T) {
+	poller := Poller{}
+
+	type test struct {
+		name string
+		out  string
+		ping float32
+		isOK bool
+		want int
```
Hardikl, line 13: want not needed.

Hardikl (approved):

cmd/poller/poller.go
```diff
@@ -504,13 +507,19 @@ func (p *Poller) handleSignals(signalChannel chan os.Signal) {
...
-				if p, err := strconv.ParseFloat(y[0], 32); err == nil {
-					return float32(p), true
-				}
+func (p *Poller) parsePing(out string) (float32, bool) {
+	if strings.Contains(out, "min/avg/max") {
+		match := pingRegex.FindStringSubmatch(out)
+		if len(match) > 0 {
+			if p, err := strconv.ParseFloat(match[1], 32); err == nil {
+				return float32(p), true
 			}
```
rahulguptajss, line 523: may be do all this in ping command itself using grep? That way no need to do golang parsing.

cgrinds: not well supported on busybox and will be more brittle

PR #569, opened 2021-10-13
feat: add Prometheus HTTP SD node

cgrinds:
- added a Prometheus HTTP service discovery end-point
- updated the documentation that motivates this feature
- updated harvest.yml with admin node info
- simplified some of the config code by removing pointers - also less code and easier to read
- removed a node-based getter that didn't merge correctly if called out of order
- removed some cmd.Exec code that is replaceable with gopsutil
- added `bin/harvest admin start` command

### DIFF
--- a/cmd/admin/admin.go
+++ b/cmd/admin/admin.go
@@ -167,12 +167,16 @@ func (a *Admin) findPrometheusPollers() {
 }
 
 func Cmd() *cobra.Command {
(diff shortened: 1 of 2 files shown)


rahulguptajss:
multiple PRs for same?
https://github.com/NetApp/harvest/pull/556
https://github.com/NetApp/harvest/pull/569

PR #626, opened 2021-11-01
refactor: clean-up conf

cgrinds:


pkg/conf/conf.go
```diff
@@ -302,66 +305,65 @@ func (p *Poller) Union(defaults *Poller) {
...
-		if defaultNode != nil {
-			password := defaultNode.GetChildContentS("password")
-			defaultStruct := *Config.Defaults
-			if defaultStruct.Password != "" && defaultStruct.Password != password {
-				defaultNode.SetChildContentS("password", defaultStruct.Password)
-			}
-		}
+// ZapiPoller creates a poller out of a node, this is a bridge between the node and struct-based code
+// Used by ZAPI based code
+func ZapiPoller(n *node.Node) Poller {
```
rahulguptajss, line 310: There are other fields documented which aren't converted in this.

labels
log_max_bytes
log_max_files
log
batch_size
schedule

cgrinds: those all should still work - this change has no impact on existing keys. It only carves out the subset of configuration that is needed to create the zapi poller

rahulguptajss (approved):

PR #640, opened 2021-11-03
fix: Harvest should specify a default datasource

cgrinds:


cdurai-netapp (dismissed):

rahulguptajss (approved):
influx dashboard also needs same change.

cgrinds:
> influx dashboard also needs same change.
Yes, separate PR

PR #645, opened 2021-11-05
feat: env var `HARVEST_CONFIG` has higher precedence than --config cm…

cgrinds:
…dline arg

### DIFF
--- a/cmd/tools/doctor/doctor.go
+++ b/cmd/tools/doctor/doctor.go
@@ -72,7 +72,7 @@ func doMerge(path1 string, path2 string) {
 
 func doDoctorCmd(cmd *cobra.Command, _ []string) {
 	var config = cmd.Root().PersistentFlags().Lookup("config")
-	doDoctor(config.Value.String())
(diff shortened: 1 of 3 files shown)


cdurai-netapp (approved):

rahulguptajss (commented):
do we need to change for admin also?

https://github.com/NetApp/harvest/blob/635791b17330aec4d71ed75345972e2f6adf107c/cmd/admin/admin.go#L223

cgrinds:
> do we need to change for admin also?
> 
> https://github.com/NetApp/harvest/blob/635791b17330aec4d71ed75345972e2f6adf107c/cmd/admin/admin.go#L223

`admin.go` and all other cmds should be covered by the one change in `conf`. The two special cases for `doctor` and `generate` are because:

- `doctor` doesn't use `conf` since it validates the yaml file directly
- `generate` needs to write a path to the config file, in addition to normal reads

rahulguptajss (approved):

PR #657, opened 2021-11-09
refactor: remove value_mapping

cgrinds:
`value_mapping` was deprecated in 21.11 and removed in next release

### DIFF
--- a/pkg/test/mergeTemplates/21.08.0_lun_merge_21.08.0_extended.yaml
+++ b/pkg/test/mergeTemplates/21.08.0_lun_merge_21.08.0_extended.yaml
@@ -14,9 +14,6 @@ counters:
     - ^vserver => svm
 plugins:
   LabelAgent:
-    value_mapping:
-      - status state online `1`
-      - custom_status state online `1`
     value_to_num:
       - new_status state online online `0`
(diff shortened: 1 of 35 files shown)


rahulguptajss (approved):

pkg/test/mergeTemplates/21.08.0_lun_merge_21.08.0_extended.yaml
```diff
@@ -14,9 +14,6 @@ counters:
     - ^vserver => svm
 plugins:
   LabelAgent:
-    value_mapping:
-      - status state online `1`
-      - custom_status state online `1`
     value_to_num:
       - new_status state online online `0`
```
Hardikl, line 18: should we add custom line 19 here also?

Hardikl (approved):

pkg/test/mergeTemplates/21.08.0_lun_merge_21.08.0_extended.yaml
cgrinds: good catch @Hardikl  - in this case, that's test code that is still correctly handled

PR #668, opened 2021-11-15
docs: document how ZAPIs, templates, and exporting work

cgrinds:


Hardikl (dismissed):

docs/TemplatesAndMetrics.md
```diff
@@ -0,0 +1,306 @@
...
+```
+# default.yaml
+objects:
+  Disk:  disk.yaml
+```
+
+- Line `2` is the name of the ZAPI that Harvest will send to collect disk resources
+- Line `3` is the prefix used to export metrics associated with this object. i.e. all metrics will be of the form `disk_*`
+- Line `5` the [counter section](https://github.com/NetApp/harvest/tree/main/conf#counters) is where we define the metrics, labels, and what constitutes instance uniqueness
+- Line `7` the double hat prefix `^^` means this attribute is an instance key used to determine uniqueness. Uuids are good choices for uniqueness
```
rahulguptajss, line 160: ^^ also adds key as label

docs/TemplatesAndMetrics.md
```diff
@@ -0,0 +1,306 @@
...
+10	      - bytes-per-sector        => bytes_per_sector        # notice this has the same hierarchical path we saw from bin/zapi
+11	      - capacity-sectors        => sectors
+12	      - ^disk-type              => type
+13	      - ^is-shared              => shared
+14	      - ^model                  => model
+15	      - ^serial-number          => serial_number
+16	      - ^shelf                  => shelf
+17	      - ^shelf-bay              => shelf_bay
+18	    - disk-ownership-info:
+19	      - ^home-node-name         => node
```
rahulguptajss, line 186: it is obvious but may be worth adding that a string value can not be a metric.

rahulguptajss (approved):

PR #678, opened 2021-11-18
feat: admin with docker compose

rahulguptajss:
Opening as draft PR as there are couple of issues which needs to be discussed/solved.

Example compose file

```yaml
version: "3.7"

services:
  admin:
    image: rahulguptajss/harvest:latest
    container_name: admin
    restart: unless-stopped
    ports:
      - 8887:8887
    entrypoint: ["bin/harvest", "admin", "start", "--config", "/opt/harvest.yml"]
    volumes:
      - type: bind
        source: /Users/rahulg2/GolandProjects/harvest/harvest.yml
        target: /opt/harvest.yml

  umeng-aff300:
    image: rahulguptajss/harvest:latest
    container_name: poller-umeng-aff300
    restart: unless-stopped
    ports:
      - 12990:12990
    command: '--poller umeng_aff300 --promPort 12990 --config /opt/harvest.yml'
    volumes:
      - type: bind
        source: /Users/rahulg2/GolandProjects/harvest/conf
        target: /opt/harvest/conf
      - type: bind
        source: /Users/rahulg2/GolandProjects/harvest/harvest.yml
        target: /opt/harvest.yml
```

1. nerdctl doesn't like new compose file. docker cli works fine.
```FATA[0000] service admin: specifying entrypoint with multiple strings ([bin/harvest admin start --config /opt/harvest.yml]) is not supported yet```

2. harvest.yaml needs to have below configuration for pollers to read ```admin``` service.

```yaml
Admin:
  httpsd:
    listen: admin:8880
```

3. httpsd end point has below json. These IPs should be visible inside docker network. Should these IP's be by service name? I think if we know pollers are running in docker config, we can adjust usecase 2 and 3 accordingly.

```json
[{"targets":["192.168.176.4:12990"],"labels":{"__meta_poller":"umeng_aff300"}},{"targets":["192.168.176.3:12991"],"labels":{"__meta_poller":"F2240-127-26"}}]```

### CODE REVIEW

**Review by Hardikl — APPROVED** [review]

File: cmd/poller/poller.go [hunk 1]
```diff
@@ -903,10 +903,12 @@ func (p *Poller) publishDetails() {
 		if exp.Type != "Prometheus" {
 			continue
 		}
-		if exp.LocalHttpAddr == "localhost" || exp.LocalHttpAddr == "127.0.0.1" {
+		if exp.LocalHttpAddr == "0.0.0.0" {
+			exporterIp = localIp
+		} else if exp.LocalHttpAddr != "localhost" || exp.LocalHttpAddr == "127.0.0.1" {
```
cgrinds, line 908: `else if exp.LocalHttpAddr == "localhost" || exp.LocalHttpAddr == "127.0.0.1"`

cmd/tools/generate/generate.go
```diff
@@ -157,10 +169,34 @@ func generateDocker(path string, kind int) {
...
-			"Start containers with:\n"+
-				color.Colorize("docker-compose -f "+opts.outputPath+" up -d --remove-orphans\n", color.Green))
+		if opts.admin {
+			httpsd := conf.Config.Admin.Httpsd.Listen
+
+			if httpsd == "" {
+				panic("Missing admin configuration in harvest configuration")
+			}
+			adminPort := 8887
+			if strings.Contains(httpsd, ":") {
```
cgrinds, line 181: Maybe a little clearer
```
if s := strings.Split(httpsd, ":"); len(s) == 2 {
    adminPort, err = strconv.Atoi(s[1])
}
```

cmd/tools/generate/generate.go
```diff
@@ -157,10 +169,34 @@ func generateDocker(path string, kind int) {
...
+				h := strings.Split(httpsd, ":")
+				adminPort, err = strconv.Atoi(h[len(h)-1])
+				if err != nil {
+					panic(err)
```
cgrinds, line 185: seems like the message logged could be the same as line 188?

cmd/poller/poller.go
rahulguptajss: Thanks. Yes this was failing the test hence i left PR in draft. Corrected it now.

cmd/tools/generate/generate.go
rahulguptajss: done

cmd/tools/generate/generate.go
rahulguptajss: done.

Hardikl (approved):

cgrinds (approved):

Issue #587, opened 2021-10-20
TopN should return no more than N resources


cgrinds:
Volume dashboard is a good way to reproduce
Commit: `702d351`

If you select top 5, Prometheus returns more than 5 resources to Grafana. Looks like a [known issue](https://stackoverflow.com/questions/38783424/prometheus-topk-returns-more-results-than-expected) with a [workaround](https://www.robustperception.io/graph-top-n-time-series-in-grafana).

- https://grafana.com/docs/grafana/latest/datasources/prometheus/ has more details

Hardikl:
**Issue**: In case of range vectors, topk function would give N records at each time-stamp basis, not at consolidated to whole given range, which result in having more than N records. For Instant vector case, topk would be working as expected and give exact N records.

topN investigation so far:
- We haven't found any grafana dashboard from catalog which has topk var, P95/P98 var is available, but it's direct query which gives the value
- Panel level variable is not supported in grafana 
- We worked together and identified the query which gives top n volume names, but join is not possible in Prometheus
- Then, we tried to join range with instant vector in grafana, but [range (op) instant = range] turn into range, which is same as more than N records, and not solving the issue in grafana
- One option is to pass this identified top n volume names in range series - which is same as introducing new global variable for each query - which led us to have many hidden variables, same as PR #708
- There are few known issues and links which tells that VictoriaMetrics supports the same, not PromQL. And few link would still point to workaround which is in PR.
https://stackoverflow.com/questions/66016655/can-you-apply-a-limit-on-promql-query
https://github.com/prometheus/prometheus/issues/586
https://stackoverflow.com/questions/43840372/how-to-show-top-10-values-only-in-grafana

**workaround**: https://www.robustperception.io/graph-top-n-time-series-in-grafana

Thanks @rahulguptajss and @cgrinds for the support.

cgrinds:
@Hardikl updated the wiki with workaround info, closing

PR #680, opened 2021-11-18
fix: network detail dashboard issues

Hardikl:
Issues: [with fix]
- all Variables have sorting disabled
--> All variables(cluster, node, Eth, Fcp) would support sorting now but Datacenter
- Ethernet Receive & Ethernet Send have Value Mappings defined that makes the timeseries show "UP" when there is 0 traffic
--> Removed the value mapping as it's not required here, When there is 0 traffic, it would show 0 B/s
- Port Utilization uses wrong unit, Percent 0-100 instead 0-1
--> Corrected the percent issue 
- Top Resources variable is not used
--> It's used for 8 graphs, Ethernet drilldown: send throughput, receive throughput, FC drildown:  send throughput, receive throughput, send latency, receive latency, link down, link failure
--> There is identified issue for TopN resource https://github.com/NetApp/harvest/issues/587, we are working in this
- Ethernet drilldown panels do not use Eth variable
--> Added the Eth var in queries
- FC panels do not use the FCP variable
--> Added the FCP var in queries
- Speed is shown in Gbyte instead of Gbit
--> converted to Gbit/s

### DIFF
--- a/grafana/dashboards/7mode/harvest_dashboard_network7.json
+++ b/grafana/dashboards/7mode/harvest_dashboard_network7.json
@@ -217,19 +217,6 @@
       "description": "",
       "fieldConfig": {
         "defaults": {
-          "mappings": [
-            {
-              "options": {
-                "0": {
-                  "text": "UP"
-                },
-                "1": {
-                  "text": "DOWN"
-                }
-              },
-              "type": "value"
-            }
-          ],
           "noValue": "0",
           "thresholds": {
             "mode": "absolute",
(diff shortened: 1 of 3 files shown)


Hardikl:
- nic.go plugin reverted and added one comment line there.
- c-mode and 7mode network detail has changed
- FCP drilldown speed still show as Bps as FCP plugin don't change any customization there and all metrics are in bytes as per the plugin usage.
- There is network dashboard in cmode having the same issues, but not changing there as it will removed in next release as we remove duplicated dashboards.

grafana/dashboards/7mode/harvest_dashboard_network7.json
```diff
@@ -816,6 +790,30 @@
                 "value": 398
               }
             ]
+          },
```
cgrinds, line 793: `calculateField` is missing from 7-mode dashboard

cgrinds (commented):
nice job!

grafana/dashboards/7mode/harvest_dashboard_network7.json
Hardikl: added

Hardikl:
it was showing earlier: 
image
image


Now, it show as :
image
image

cgrinds (approved):

PR #699, opened 2021-11-25
fix: remove duplicate dashboards: network, node, svm, volume

Hardikl:
Summary:

Network: --> Mentioned issue is due to order of transformation applied, not due to duplication. --> Removal Done
Ethernet ports table: send and receive guage correction in detail
Rest all are fine and detail would cover the fcp nvmf as well

Node: --> No changes needed --> Removal Done
Avg latency: detail is using avg(volume_avg_latency{})/1000 with unit ms where as node page use avg(volume_avg_latency{}) with unit us
Rest all fine and detail would cover all

SVM: --> copied 2 graphs from non-detail page to detail page[NFSv3 Latency by Op Type, NFSv3 IOPs per Type] --> Removal Done
Latency per type: avg by __name__ in svm not in detail
Same as iops per type
Avg throughput, avg iops table not available in detail(due to sync with v3, v4 and v4.1)
Rest all fine and detail would cover all

Volume: --> Removal Done
All fine and detail would cover all


rahulguptajss:
Could you open an git issue and track all which needs fixing so that we can test that area later.

rahulguptajss (approved):

cgrinds (approved):

cgrinds:
Thanks @Hardikl

cgrinds:
Verified on `release/22.02.0` bae2dc8

PR #702, opened 2021-11-26
fix: handle booleans when merging

### LINKED ISSUE #701: UseInsecureTls should not be overridden when defined
Discovered on `21.11`

This config:

```
Defaults:
  use_insecure_tls: true
Pollers:
  umeng_aff300:
    use_insecure_tls: false
```

When merged, becomes:

```
Pollers:
  umeng_aff300:
    use_insecure_tls: true
```

cgrinds:
Fixes #701


rahulguptajss:
Doesn't seem to have any file in PR

PR #708, opened 2021-11-29
fix: handle topk: volume POC

Hardikl:
Approach detail:
Support topk when filtered based on global variable. 

issue: topk would by default find top k element from different different time range and show all, it's not able to choose from consolidated list.

reference: https://www.robustperception.io/graph-top-n-time-series-in-grafana

Fix: 
- when filtered based on svm, it gets all volumes from svm with topk applied
- when filtered based on volume, it gets only volume records, not topk
- when not filtered on svm/volume, it gets records with topk applied
 
There are few disadvantages of this approach:
1. too many hidden variables in dashboard
image

2. If topk query identify volume VOL1, then it list all volumes named VOL1 from all svms even it's not belongs to top most list. [This is same as volume filter, it shows all volume with same name when applied]

### DIFF
--- a/grafana/dashboards/cmode/harvest_dashboard_volume_details.json
+++ b/grafana/dashboards/cmode/harvest_dashboard_volume_details.json
@@ -129,7 +129,7 @@
       "targets": [
         {
           "exemplar": false,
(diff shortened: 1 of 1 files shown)


rahulguptajss:
Can this be closed?

PR #719, opened 2021-12-02
fix: include missing server_cert.cnf

cgrinds:


docs/samples/server_cert.cnf
```diff
@@ -0,0 +1,21 @@
...
+OU  = Harvest
+CN  = /
+
+[req_ext]
+subjectAltName = @alt_names
+
+[alt_names]
+DNS.1                   = umeng-aff300-05-06-cm.rtp.openenglab.netapp.com
+DNS.2                   = umeng-aff300-05-06
+IP.1                    = 10.193.48.11
```
Hardikl, line 20: customer would change this i/p based on their network, right ?

Hardikl (approved):

docs/samples/server_cert.cnf
cgrinds: that's right, the instructions mention that

PR #735, opened 2021-12-09
feat: include plugin label agent

rahulguptajss:


cmd/poller/plugin/README.md
```diff
@@ -120,6 +120,9 @@ Notice that the rules are executed in the same order as you've added them. List
   - [exclude_equals](#exclude_equals)
   - [exclude_contains](#exclude_contains)
   - [exclude_regex](#exclude_regex)
+  - [include_equals](#include_equals)
```
Hardikl, line 123: We should add in readme that include and exclude should not apply on the same label OR order matters like below. 
As exclude applicable for given record, but include would change for whole records.

Example:
exclude_equals: style `flexgroup_constituent`.  [flexvol and flexgroup comes in result]
include_equals: style `flexgroup`.                        [flexvols removed]


include_equals: style `flexgroup`.                        [only flexgroup comes in result]
exclude_equals: style `flexgroup_constituent`.  [No effect]

cmd/poller/plugin/label_agent/label_agent.go
```diff
@@ -192,7 +195,40 @@ func (me *LabelAgent) excludeRegex(instance *matrix.Instance) {
...
+		}
+	}
+}
+
+// if label is not equal to value, set instance as non-exportable
+func (me *LabelAgent) includeEquals(instance *matrix.Instance) {
+	for _, r := range me.includeEqualsRules {
+		if instance.GetLabel(r.label) != r.value {
+			instance.SetExportable(false)
+			me.Logger.Trace().Msgf("includeEquals: (%s) [%s] instance with labels [%s] => excluded", r.label, r.value, instance.GetLabels().String())
```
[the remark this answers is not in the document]
rahulguptajss: this logic is to exclude only.

cmd/poller/plugin/README.md
rahulguptajss: Not sure if we should document such cases as it will be implicit by rule defintion in docs.

cmd/poller/plugin/label_agent/parse_rules.go
```diff
@@ -375,6 +396,63 @@ func (me *LabelAgent) parseExcludeRegexRule(rule string) {
...
+func (me *LabelAgent) parseIncludeRegexRule(rule string) {
+	if fields := strings.SplitN(rule, " `", 2); len(fields) == 2 {
+		r := includeRegexRule{label: fields[0]}
+		var err error
+		if r.reg, err = regexp.Compile(strings.TrimSuffix(fields[1], "`")); err == nil {
+			me.includeRegexRules = append(me.includeRegexRules, r)
+			me.Logger.Debug().Msgf("(include_regex) compiled regex: [%s]", r.reg.String())
+			me.Logger.Debug().Msgf("(include_regex) parsed rule [%v]", r)
+		} else {
+			me.Logger.Error().Stack().Err(err).Msgf("(include_regex) compile regex:")
```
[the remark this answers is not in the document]
rahulguptajss: Could you explain your comment?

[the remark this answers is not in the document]
Hardikl: ohh, I missed to see err in log. 
pls ignore.

Hardikl (dismissed):

cmd/poller/plugin/label_agent/label_agent.go
```diff
@@ -192,12 +201,63 @@ func (me *LabelAgent) excludeRegex(instance *matrix.Instance) {
...
+				Str("value", r.value).
+				Str("instance labels", instance.GetLabels().String()).
+				Msg("includeContains: included")
+			break
+		}
+	}
+	instance.SetExportable(isExport)
+}
+
+// if label does not equal to value, set instance as non-exportable
```
cgrinds, line 245: // if label does not match regex, do not export the instance
or with fewer negatives
// only export instances with a matching (regex) label
// if an instance does not match the regex label it will not be exported

cmd/poller/plugin/README.md
```diff
@@ -287,18 +290,74 @@ Same as `exclude_equals`, but will use a regular expression and all matching ins
...
 ```yaml
-exclude_equals: vol_type `flexgroup_`
+exclude_regex: vol_type `^flex`
+# all instances, which have label "vol_type" which starts with
+# "flex" will not be exported
+```
+
+## include_equals
+
+Include each instance, if the value of `LABEL` is exactly `VALUE`. Include means that metrics for this instance will be exported.
```
cgrinds, line 307: Include means that metrics for this instance will be exported and instances that do not match will not be exported.

cmd/poller/plugin/README.md
```diff
@@ -287,18 +290,74 @@ Same as `exclude_equals`, but will use a regular expression and all matching ins
...
+```yaml
+include_contains: vol_type `flexgroup_`
 # all instances, which have label "vol_type" which contain
-# "flexgroup_" will not be exported
+# "flexgroup_" will be exported
+```
+
+## include_regex
+
+Same as `include_equals`, but will use a regular expression and all matching instances will be included.
```
cgrinds, line 345: Same as `include_equals`, but a regular expression will be used for inclusion. Similar to the other includes, all matching instances will be included and all non-matching will not be exported.

PR #779, opened 2022-01-12
feat: add port rest template

rahulguptajss:


conf/rest/9.13.0/snapshot.yaml
```diff
@@ -2,10 +2,6 @@ name:                     Snapshot
 query:                    api/private/cli/snapshot/policy
 object:                   snapshot
 
-schedule:
```
cgrinds, line 5: Unrelated change?

cgrinds:
Looks good, one minor comment

cgrinds (approved):

conf/rest/9.13.0/snapshot.yaml
rahulguptajss: yes added in this PR based on hardik's changes in https://github.com/NetApp/harvest/pull/778

PR #783, opened 2022-01-17
feat: rest framework enhancements

rahulguptajss:


cmd/collectors/rest/rest.go
```diff
@@ -337,12 +342,26 @@ func (r *Rest) PollData() (*matrix.Matrix, error) {
...
-		for key, metric := range r.Matrix.GetMetrics() {
-			f := instanceData.Get(key)
+		for _, metric := range r.prop.metrics {
+			f := instanceData.Get(metric.name)
 			if f.Exists() {
-				if err = metric.SetValueFloat64(instance, f.Float()); err != nil {
-					r.Logger.Error().Err(err).Str("key", key).Str("metric", metric.GetName()).
-						Msg("Unable to set float key on metric")
+				if metr, ok := r.Matrix.GetMetrics()[metric.name]; !ok {
+					if metr, _ = r.Matrix.NewMetricFloat64(metric.name); err != nil {
```
cgrinds, line 349: would be slightly clearer to 1) create the metric by name and 2) update the map in this block and then remove the else and use it for both cases

cmd/collectors/rest/rest.go
```diff
@@ -464,23 +483,23 @@ func (r *Rest) processEndPoints(data *matrix.Matrix) error {
 				}
 
 				for _, metric := range endpoint.prop.metrics {
-					f := instanceData.Get(metric)
+					f := instanceData.Get(metric.name)
 					if f.Exists() {
-						if metr, ok := data.GetMetrics()[metric]; !ok {
-							if metr, _ = data.NewMetricFloat64(metric); err != nil {
+						if metr, ok := data.GetMetrics()[metric.name]; !ok {
```
cgrinds, line 488: similar comment as above about minor refactoring

cmd/collectors/rest/rest.go
Hardikl, line 349: err not used

rahulguptajss: done

cmd/collectors/rest/rest.go
rahulguptajss: done

cmd/collectors/rest/rest.go
rahulguptajss: done

cmd/collectors/rest/rest.go
```diff
@@ -337,11 +342,20 @@ func (r *Rest) PollData() (*matrix.Matrix, error) {
...
+		for _, metric := range r.prop.metrics {
+			metr, ok := r.Matrix.GetMetrics()[metric.name]
+			if !ok {
+				if metr, err = r.Matrix.NewMetricFloat64(metric.name); err != nil {
+					r.Logger.Error().Err(err).
+						Str("name", metric.name).
+						Msg("NewMetricFloat64")
+				}
+			}
+			f := instanceData.Get(metric.name)
```
cgrinds, line 354: when err != nil is metr valid to use below?

cmd/collectors/rest/rest.go
```diff
@@ -464,25 +478,20 @@ func (r *Rest) processEndPoints(data *matrix.Matrix) error {
...
-					f := instanceData.Get(metric)
+					metr, ok := data.GetMetrics()[metric.name]
+					if !ok {
+						if metr, err = data.NewMetricFloat64(metric.name); err != nil {
+							r.Logger.Error().Err(err).
+								Str("name", metric.name).
+								Msg("NewMetricFloat64")
+						}
+					}
+					f := instanceData.Get(metric.name)
```
cgrinds, line 489: when err != nil, is metr valid?

cmd/collectors/rest/templating.go
```diff
@@ -28,16 +27,15 @@ func (r *Rest) initCache() error {
...
-		err                 error
 	)
 
-	if x := r.Params.GetChildContentS("object"); x != "" {
-		r.Object = x
-	} else {
-		r.Object = strings.ToLower(r.Object)
+	if r.Object == "" {
+		if x := r.Params.GetChildContentS("object"); x != "" {
+			r.Object = x
```
cgrinds, line 34: are we certain that x has already been lowercased?

rahulguptajss: we are lower casing it at line number 38 anyways

cmd/collectors/rest/rest.go
rahulguptajss: metr is not added to matrix if there is an error. In current case, we check for duplicates.

PR #794, opened 2022-01-21
feat: task recovery handling

### LINKED ISSUE #480: PollData task Recovery handling is not correct
When cluster connection is lost or if there are no instances, "data" task during ZapiPerf collector recovers all other tasks. This causes log spam like below and unnecssary early retry to cluster.

**Target unreachable case:**
7:23PM WRN goharvest2/cmd/poller/collector/collector.go:313 > target unreachable, entering standby mode (retry in retryDelay s) Poller=F2240-127-26 collector=ZapiPerf:CIFSvserver retryDelay=4 task=counter
7:23PM INF goharvest2/cmd/poller/collector/collector.go:355 > recovered from standby mode, back to normal schedule Poller=F2240-127-26 collector=ZapiPerf:CIFSvserver task=data

**No instance found case:**
{"level":"info","Poller":"umeng_aff300","collector":"ZapiPerf:Path","caller":"goharvest2/cmd/poller/collector/collector.go:323","time":"2021-08-25T18:10:16+05:30","message":"task instance, no [Path] instances on system, entering standby mode"}
{"level":"info","Poller":"umeng_aff300","collector":"ZapiPerf:Path","caller":"goharvest2/cmd/poller/collector/collector.go:346","time":"2021-08-25T18:10:16+05:30","message":"task data, recovered from standby mode, back to normal schedule"}

rahulguptajss:
Fixes #480

### DIFF
--- a/cmd/poller/collector/collector.go
+++ b/cmd/poller/collector/collector.go
@@ -281,7 +281,7 @@ func (me *AbstractCollector) Start(wg *sync.WaitGroup) {
 				continue
 			}
 
-			if !me.Schedule.IsTaskStandBy(task) {
+			if me.Schedule.IsStandBy() && !me.Schedule.IsTaskStandBy(task) {
 				me.Logger.Info().
 					Str("task", task.Name).
 					Msgf("schedule is in standby mode skipping")
@@ -424,7 +424,7 @@ func (me *AbstractCollector) Start(wg *sync.WaitGroup) {
 			me.Schedule.Sleep()
 			// log if lagging by more than 50 ms
 			// < is used since larger durations are more negative
-		} else if nd.Milliseconds() <= -50 {
+		} else if nd.Milliseconds() <= -50 && !me.Schedule.IsStandBy() {
 			me.Logger.Warn().Msgf("lagging behind schedule %s", (-nd).String())
 		}
 	}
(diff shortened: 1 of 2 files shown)


cmd/poller/collector/collector.go
```diff
@@ -281,6 +281,13 @@ func (me *AbstractCollector) Start(wg *sync.WaitGroup) {
 				continue
 			}
 
+			if me.Schedule.IsStandBy() && !me.Schedule.IsTaskStandBy(task) {
```
cgrinds, line 284: IsTaskStandBy makes the same check, so we could simplify this to 
```go
if !me.Schedule.IsTaskStandBy(task) {
```
or we could flip it around to this which maybe reads a little better?

`if me.Schedule.ShouldSkip(task) {`

cgrinds, line 284: minor: `Msg` instead of `Msgf` since no substitution

rahulguptajss: Both are different. Is !TaskStandBy will be true in all happy cases. Hence added additional standbycheck outside

rahulguptajss: done

rahulguptajss: removed an redundant check from IsTaskStandBy.

cgrinds (approved):

PR #806, opened 2022-01-27
docs: add rest-collector architecture doc

cgrinds:


rahulguptajss (approved):

florianmulatz:
What do you mean by 9.12+ ? - do you mean 9.9P12. Not clearly sure for me. Please clarify

florianmulatz:
I was confused since there is no 9.12 currently. Overread that this is meant to be for future 9.12 release! Thx!

rahulguptajss:
yes, Rest collector is officially for 9.12+ only. If you want to try this out for a use-case which is only available through Rest in pre 9.12 versions, we can help you in setting that up with Rest Collector.

PR #822, opened 2022-02-03
fix: pin Grafana and Prometheus versions in compose workflow

cgrinds:


cmd/tools/grafana/grafana.go
```diff
@@ -456,6 +456,7 @@ func importFiles(dir string, folder Folder) {
 		request = make(map[string]interface{})
 		request["overwrite"] = true
 		request["folderId"] = folder.folderId
+		//delete(dashboard, "id")
```
Hardikl, line 459: non related

Hardikl (approved):

rahulguptajss (approved):

PR #825, opened 2022-02-04
fix: id issue in grafana import/export

Hardikl:
fix for dashboard import when dashboard id is not null.

### DIFF
--- a/cmd/tools/grafana/grafana.go
+++ b/cmd/tools/grafana/grafana.go
@@ -430,7 +430,7 @@ func importFiles(dir string, folder Folder) {
 			continue
 		}
 
-		// Updating the id of dashboards, in case id is not null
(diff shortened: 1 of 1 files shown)


rahulguptajss (dismissed):

cmd/tools/grafana/grafana.go
```diff
@@ -430,6 +430,15 @@ func importFiles(dir string, folder Folder) {
 			continue
 		}
 
+		// Updating the id of dashboards, in case id is not null
```
cgrinds, line 433: Might be clearer to say something like > If the dashboard has an id defined, change the id to empty string so Grafana treats this as a new dashboard instead of an update to an existing one

rahulguptajss (approved):

PR #845, opened 2022-02-10
feat: rest gaps

rahulguptajss:


conf/rest/9.12.0/disk.yaml
```diff
@@ -30,6 +29,7 @@ endpoints:
       - ^^uid => uid
       - sectors_read => stats_sectors_read
       - sectors_written => stats_sectors_written
+      - disk_io_kbps_total => stats_io_kbps
```
rahulguptajss, line 32: stats.throughput is in bytes in public rest. We'll need plugin to covert to kbps. We can do that later.

conf/rest/9.12.0/node.yaml
```diff
@@ -28,6 +24,7 @@ endpoints:
       - ^max_aggr_size => max_aggr_size
       - ^max_node_vvols => max_vol_num
       - ^max_vol_size => max_vol_size
+      - ^health => healthy
```
rahulguptajss, line 27: state is coming as `up` in public rest. As per zapi, we need it as `true`

conf/rest/9.12.0/node.yaml
```diff
@@ -8,17 +8,13 @@ counters:
...
   - ^controller.failed_power_supply.message.message => failed_power_message
   - ^controller.over_temperature => over_temperature
-  - ^state => healthy
   - ^location
   - ^model
   - ^serial_number => serial
   - ^version.full => version
   - uptime
   - controller.failed_power_supply.count => failed_power
-  - metric.processor_utilization => cpu_busytime   #hidden field
```
rahulguptajss, line 18: value is quite different and an rejected gap. Removing for now.

Hardikl (approved):

PR #877, opened 2022-02-23
feat: security infra as collector

Hardikl:
image

image

image

Note: 
all metrics in this POC are ends with c just for easy to test and referencing.


Hardikl:
As discussed, we will not follow this approach. Closing this PR.

PR #894, opened 2022-03-10
feat: restperf templates

rahulguptajss:


conf/restperf/default.yaml
```diff
@@ -11,13 +11,17 @@ objects:
   CIFSNode:         cifs_node.yaml
   HeadroomCPU:      resource_headroom_cpu.yaml
   HeadroomAggr:     resource_headroom_aggr.yaml
-  Qtree:            qtree.yaml
   NFSv3Node:        nfsv3_node.yaml
+  NFSv4Node:        nfsv4_node.yaml
+  NFSv41Node:       nfsv4_1_node.yaml
+#  VolumeNode:       volume_node.yaml
+  WAFL:             wafl.yaml
+  Qtree:            qtree.yaml
```
Hardikl, line 19: If we want to follow ascending order then we can interchange.

Hardikl (approved):

conf/restperf/default.yaml
rahulguptajss: yes. Done

cgrinds (approved):

PR #896, opened 2022-03-10
feat: rest gaps

rahulguptajss:


conf/zapi/cdot/9.8.0/disk.yaml
```diff
@@ -14,7 +14,7 @@ counters:
       - ^is-shared              => shared
       - ^model                  => model
       - ^serial-number          => serial_number
-      - ^shelf                  => shelf
+      - ^shelf-uid              => shelf
```
rahulguptajss, line 17: zapi gives always zero. So mapped with uid same as of rest.

Hardikl (approved):

PR #924, opened 2022-03-25
ci: use pretty logger instead of json

cgrinds:


integration/test/native_installer_test.go
```diff
@@ -1,4 +1,5 @@
-//+build install_native
+//go:build install_native
```
rahulguptajss, line 1: adding them back? they were just removed?

cgrinds: yep, I planned on merging into this one after CI runs

PR #928, opened 2022-03-28
docs: object_store_client_op template is being deprecated

cgrinds:


rahulguptajss (dismissed):

ruanruijuan (commented):
I think you meant this ZapiPerf template will be deprecated for ZAPI collector, but the reason is because ONTAP is deprecating the support in REST. It's not clear enough of the ZAPI deprecation, which means the deprecation of this object essentially.

PR #936, opened 2022-03-30
feat: credentials_file should support defaults

cgrinds:


pkg/conf/conf.go
```diff
@@ -108,7 +108,12 @@ func ReadCredentialsFile(credPath string, p *Poller) error {
 
 	credPoller := credConfig.Pollers[p.Name]
 	if credPoller == nil {
-		return errors.New(errors.INVALID_PARAM, "poller not found in credentials file")
+		// when the poller is not listed in the file, check if there is a default, and if so, use it
+		if credConfig.Defaults != nil {
+			credPoller = credConfig.Defaults
```
rahulguptajss, line 113: may be a log will help if we are using default config

rahulguptajss (dismissed):

rahulguptajss (approved):

PR #953, opened 2022-04-06
feat: power dashboard rest implementation

rahulguptajss:


conf/rest/9.12.0/shelf.yaml
```diff
@@ -30,7 +30,6 @@ plugins:
 
 export_options:
   instance_keys:
-    - shelf_uid
```
Hardikl, line 33: remove line7 as well.

cmd/collectors/rest/plugins/shelf/shelf.go
```diff
@@ -296,5 +306,154 @@ func (my *Shelf) Run(data *matrix.Matrix) ([]*matrix.Matrix, error) {
 		return true
 	})
 
+	return my.calculateEnvironmentMetrics(output, data)
+}
+
+func (my *Shelf) calculateEnvironmentMetrics(output []*matrix.Matrix, data *matrix.Matrix) ([]*matrix.Matrix, error) {
```
Hardikl, line 312: output var hasn't been used inside function, right ?

Hardikl (approved):

cmd/collectors/rest/plugins/shelf/shelf.go
rahulguptajss: not needed. corrected.

PR #977, opened 2022-04-20
fix: exclude root aggr

Hardikl:
1. aggrs before any changes

aggr_labels for rest
image

aggr labels for zapi (root_aggr label is visible, as we only show non-root aggr, so removed that label)
image

aggr metrics for both
image

2. Added the below changes to add label and exclude from label and metrics

aggr labels for zapi
image

aggr metrics for both
image


cgrinds (approved):
short and sweet 👍

PR #983, opened 2022-04-25
feat: rest perf templates

rahulguptajss:


conf/restperf/default.yaml
```diff
@@ -9,27 +9,38 @@ schedule:
 
 objects:
   CIFSNode:          cifs_node.yaml
-  Fcp:               fcp.yaml
+  Disk:              disk.yaml
+  ExtCacheObj:       ext_cache_obj.yaml
   FCVI:              fcvi.yaml
+  FcpPort:           fcp.yaml
```
Hardikl, line 15: object name FcpPort?

rahulguptajss: kept it same as zapiperf.. but i agree , we should change it to just fcp

cgrinds (approved):
nice job @rahulguptajss 👍

PR #984, opened 2022-04-26
fix: ignore transient volumes

Hardikl:
Below details from the Netapp_Harvest_IAG_1.X.pdf

****C.** Use a blacklist to prevent visibility of temporary volumes [optional]**
Harvest will discover all instances of objects such as lifs, volumes, and luns on the cluster. Some of these instances may be temporary, such as those used by backup jobs, and add undesired clutter and storage consumption. Graphite provides the ability to blacklist (block) incoming metrics that match a regular expression string and this feature can be used to block metrics for these temporary instances.
The following steps can be used to block metrics on temporary volumes that are often created by NetApp and NetApp partner software integrations:

1. Login the graphite host and if not logged in as the root superuser, become root using sudo:
```
[user@host ~]# sudo -i
```

2. Use a text editor (nano, vi) to edit the blacklist.conf file. Depending on your installation the location can vary, but is typically one of:
```
[root@host ~]# nano /etc/carbon/blacklist.conf
or
[root@host ~]# nano /opt/graphite/conf/blacklist.conf
```

3. Add the following contents to this file, save, and exit to the command prompt.
```
#
# Exclusion for SnapProtect/CommVault Intellisnap
# Clone volumes gets a “_CVclone” suffix
# ^netapp\.(capacity|perf7?)\..+\.vol\..+_CVclone\..+
#
# Exclusion for SnapCenter
# Clone volumes gets a “DDMMYYhhmmss” suffix
# ^netapp\.(capacity|perf)\..+\.vol\..+(0[1-9]|[12][0-9]|3[01])(0[1-9]|1[012])\d\d[0- 9]{6}\..+
#
# Exclusion for SnapCreator
# This is from creating a clone manually in SnapCreator
# Clone volume gets a “cl_” prefix and a “_YYYYMMDDhhmmss” suffix
#
^netapp\.(capacity|perf7?)\..+\.vol\.cl_.+_(19|20)\d\d(0[1-9]|1[012])( 0[1-9]|[12][0- 9]|3[01])[0-9]{6}\..+
#
# Exclusion for SnapDrive/SnapManager
# Clone volumes gets a “sdw_cl_” prefix
# ^netapp\.(capacity|perf7?)\..+\.vol\.sdw_cl_.+\..+
#
# Exclusion for Metadata volumes that may also clutter menus. #
# CRS volumes in SVM-DR or MetroCluster get a “MDV_CRS_” prefix
# Audit volumes get a “MDV_aud_” prefix # ^netapp\.(perf)\..+\.vol\.MDV_CRS_.+\..+ ^netapp\.(perf)\..+\.vol\.MDV_aud_.+\..+
```

**As per the 1.X doc, these are not ignored by default, They can be ignored based on the above blacklist file.**

--> No transient volumes in my local setup, but MDV_CRS and MDV_aud. Tested them with given regex.

### DIFF
--- a/conf/rest/9.12.0/volume.yaml
+++ b/conf/rest/9.12.0/volume.yaml
@@ -73,6 +73,20 @@ plugins:
         - style `flexgroup_constituent`
       replace:
         - style style `flexgroup_constituent` `flexgroup`
+      # To prevent visibility of transient volumes, uncomment the following lines
+#      exclude_regex:
+#        # Exclude SnapProtect/CommVault Intellisnap, Clone volumes have a “_CVclone” suffix
+#        - volume `.+_CVclone`
+#        # Exclude SnapCenter, Clone volumes have a “DDMMYYhhmmss” suffix
+#        - volume `.+(0[1-9]|[12][0-9]|3[01])(0[1-9]|1[012])\d\d[0-9]{6}`
+#        # Exclude manually created SnapCreator clones, Clone volumes have a “cl_” prefix and a “_YYYYMMDDhhmmss” suffix
+#        - volume `cl_.+_(19|20)\d\d(0[1-9]|1[012])( 0[1-9]|[12][0-9]|3[01])[0-9]{6}`
+#        # Exclude SnapDrive/SnapManager, Clone volumes have a “sdw_cl_” prefix
+#        - volume `sdw_cl_.+`
+#        # Exclude Metadata volumes, CRS volumes in SVM-DR or MetroCluster have a “MDV_CRS_” prefix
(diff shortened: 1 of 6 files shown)


Hardikl:
I will check perf and REST would need this filter or not and will update PR.

rahulguptajss (approved):

Hardikl:
local setup don't have all kind of transient volumes, tested with dummy volumes to make sure indentation in yaml would work as expected.

cgrinds (approved):

rahulguptajss (approved):

PR #1004, opened 2022-05-04
fix: svm Latency numbers differ significantly on Harvest 1.6 vs 2.0

### LINKED ISSUE #1003: SVM Latency numbers differ significantly on Harvest 1.6 vs Harvest 2.0
### Discussed in https://github.com/NetApp/harvest/discussions/940

<div type='discussions-op-text'>

<sup>Originally posted by **jmg011** April  1, 2022</sup>
We have some users using both harvest 1.6 dashboards and Harvest 2.0 dashboards. Some of the metrics I have set for Harvest 2.0 shows high latency at SVM level compared to harvest 1.6 metrics. 

**For example:** 

**In Harvest 1.6 query,** the SVM level latency is using vol_summary avg_latency  -> This is 0.1 ms for an svm in my test case. This matches with the "statistics vserver latency show" command on the filer

`_[netapp].[perf].[$Group].[$Cluster].[svm].[$SVM].[vol_summary].[avg_latency]_` 

**In Harvest 2.0 query,** for SVM level latency, I'm using below query. This query is reporting 25 ms latency for the same SVM. If I remove CIFS latency, it still reports almost 24 ms latency for my svm test case during the same time.  This is not matching with the "statistics vserver latency show" or "statistics volume latency show" commands on the filer. 

`avg(svm_nfs_latency{svm=~"$SVM"})  + avg(svm_cifs_latency{svm=~"$SVM"})`

I'm wondering why these two numbers differ so significantly?

**In Harvest 2.0**, when I calculate latency of all volumes on svm and sum it, it is still higher like 17ms for this test case SVM.

`sum by (svm)(volume_avg_latency{svm="testcase_svm"})/1000`

Are there different objects being used for both harvest 1.6 vs harvest 2.0 SVM latency calculations?</div>

rahulguptajss:
Fixes #1003

### DIFF
--- a/conf/restperf/default.yaml
+++ b/conf/restperf/default.yaml
@@ -43,8 +43,8 @@ objects:
   NFSv41:            nfsv4_1.yaml
   NFSv4:             nfsv4.yaml
   Volume:            volume.yaml
-  WAFLCompBin:       wafl_comp_aggr_vol_bin.yaml
   VolumeSvm:         volume_svm.yaml
+  WAFLCompBin:       wafl_comp_aggr_vol_bin.yaml
   #  Uncomment to collect workload/QOS counters
 #  Workload:             workload.yaml
 #  WorkloadDetail:       workload_detail.yaml
--- a/conf/zapiperf/default.yaml
+++ b/conf/zapiperf/default.yaml
@@ -48,8 +48,8 @@ objects:
   NFSv42:                 nfsv4_2.yaml
   NFSv4:                  nfsv4.yaml
   Volume:                 volume.yaml
-  WAFLCompBin:            wafl_comp_aggr_vol_bin.yaml
   VolumeSvm:              volume_svm.yaml
+  WAFLCompBin:            wafl_comp_aggr_vol_bin.yaml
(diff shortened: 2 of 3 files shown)


conf/restperf/default.yaml
```diff
@@ -44,7 +44,7 @@ objects:
   NFSv4:             nfsv4.yaml
   Volume:            volume.yaml
   WAFLCompBin:       wafl_comp_aggr_vol_bin.yaml
-
+  VolumeSvm:         volume_svm.yaml
```
cgrinds, line 47: maybe alphabetize?

conf/zapiperf/default.yaml
```diff
@@ -49,6 +49,7 @@ objects:
   NFSv4:                  nfsv4.yaml
   Volume:                 volume.yaml
   WAFLCompBin:            wafl_comp_aggr_vol_bin.yaml
+  VolumeSvm:              volume_svm.yaml
```
cgrinds, line 52: maybe alphabetize

cgrinds (approved):

PR #932, opened 2022-03-29
feat: power dashboard

rahulguptajss:
image
image

Issue #903, opened 2022-03-15
Harvest should include power consumed in existing shelf and cluster dashboards


cgrinds:
We already include fan speed and temperature and in a few cases power.
We should improve the power consumption metrics.

rahulguptajss:
- [ ] Zapi plugin implementation
- [ ] Rest Plugin implementation
- [ ] Dashboard changes

cgrinds:
verified power dashboard, cluster and shelf power panels in `22.05`

Issue #859, opened 2022-02-16
Harvest panics when trying to merge empty template


cgrinds:
**Describe the bug**
James Fong on Slack ran into a poller panic when running with an empty template file. The panic happens during `PreprocessTemplate`

release `22.02`

Stack follows

```
[...]

PR #1006, opened 2022-05-05
docs: changelog for 22.05

Hardikl:


ruanruijuan (commented):
- For the perf REST early access, customers should use with ONTAP 9.11.1GA. A good portion of perf REST objects are not in 9.11.1RC1 yet. But normally we do not call out a future release. Maybe a general comment of should with ONTAP version that supports perf REST?
- Power dashboard is repeated twice in the feature list (#903 and #932)
- Don’t see the mention of “vaults” support/integration
- Do we want to call out the NetApp's container registry?

ruanruijuan (commented):
One last comment: for #859, remove "the Slack". The reporter ID is on github...

ruanruijuan (dismissed):

cgrinds (approved):

PR #1009, opened 2022-05-06
docs: describe ontapi permissions required for security dashboard

cgrinds:


Hardikl (dismissed):
one same comment as Rahul has mentioned

Hardikl (approved):

PR #1012, opened 2022-05-09
fix: disk uptime to seconds in Rest

rahulguptajss:


conf/rest/9.12.0/disk.yaml
```diff
@@ -36,6 +36,8 @@ plugins:
   - LabelAgent:
       value_to_num:
         - new_status outage - - `0` #ok_value is empty value, '-' would be converted to blank while processing.
+      compute_metric:
+        - uptime MULTIPLY stats.power_on_hours 60 60
```
Hardikl, line 40: We already have one small disk rest plugin, this can be done there as well.

This approach seems changing the definition of `compute_metric`, So need to change doc here: https://github.com/NetApp/harvest/blob/33e2bc610ffa4d9bf2269cc604d8f58e5eee0904/cmd/poller/plugin/README.md#compute_metric

cgrinds (approved):

cmd/poller/plugin/README.md
```diff
@@ -481,6 +480,13 @@ compute_metric:
 # total_bytes = bytes_per_sector * sector_count
 ```
 
+```yaml
+compute_metric:
+  - uptime MULTIPLY stats.power_on_hours 60 60
+# value of metric "uptime" would be multiplication of metric value of stats.power_on_hours and scalar value of 60 * 60.
+# total_bytes = bytes_per_sector * sector_count
```
Hardikl, line 487: equation of uptime

PR #1035, opened 2022-05-17
feat: ems plugin design changes

rahulguptajss:


cmd/poller/collector/collector.go
```diff
@@ -521,6 +532,8 @@ func (me *AbstractCollector) LoadPlugins(params *node.Node, c Collector) error {
 
 	var p plugin.Plugin
 	var abc *plugin.AbstractPlugin
+	var plugins []plugin.Plugin
+	me.Plugins = make(map[string][]plugin.Plugin)
```
Hardikl, line 536: would be better to move this initialize line at 198

Hardikl: rest look good as per the meeting discussion

cmd/poller/collector/collector.go
```diff
@@ -365,21 +368,29 @@ func (me *AbstractCollector) Start(wg *sync.WaitGroup) {
 			}
 
 			if data != nil {
-				results = append(results, data)
+
+				v := make([]*matrix.Matrix, 0, len(data))
```
cgrinds, line 372: I think these lines can be simplified to
```go
for _, value := range data {
    results = append(results, value)
}
```

cgrinds (commented):
looks good @rahulguptajss

cgrinds (approved):

PR #1038, opened 2022-05-18
fix: plugin nil check

rahulguptajss:


Hardikl (approved):

cmd/poller/collector/collector.go
```diff
@@ -540,6 +540,9 @@ func (me *AbstractCollector) LoadPlugins(params *node.Node, c Collector) error {
 			p = c.LoadPlugin(name, abc)
 			me.Logger.Debug().Msgf("loaded plugin [%s]", name)
 		}
+		if p == nil {
```
cgrinds, line 543: good catch - wondering if a log message is warranted when p is nil. I haven't checked but it's likely the case that all callers of `c.LoadPlugin(name, abc)` log a msg before returning nil in which case also logging here may be redundant

cgrinds (approved):

PR #1042, opened 2022-05-19
fix(grafana): StorePool legend should read CopyStateAlloc

cgrinds:
Thanks to Rusty Brown for reporting


cgrinds:
Thanks to Rusty Brown for reporting

PR #1060, opened 2022-05-31
fix: flexgroup latency metric fix

rahulguptajss:


cgrinds (approved):

cgrinds:
As reported by PapadopoulosA on Slack. Thanks!

PR #1068, opened 2022-06-02
feat: log APIRequestRejected as warning

cgrinds:


pkg/errors/errors.go
```diff
@@ -9,6 +9,12 @@ import (
 	"strings"
 )
 
+type apiError string
+
+const APIRequestRejected = apiError("API request rejected")
+
+func (e apiError) Error() string { return string(e) }
```
Hardikl, line 16: not able to find the usage of this.

Hardikl: Got it.

PR #1071, opened 2022-06-06
feat: consolidate log messages and reduce noise

cgrinds:
feat: improve qtree plugin logging

### DIFF
--- a/cmd/collectors/rest/plugins/shelf/shelf.go
+++ b/cmd/collectors/rest/plugins/shelf/shelf.go
@@ -12,6 +12,7 @@ import (
 	"github.com/netapp/harvest/v2/pkg/tree/node"
 	"github.com/netapp/harvest/v2/pkg/util"
 	"github.com/tidwall/gjson"
+	"sort"
 	"strings"
 	"time"
 )
@@ -189,7 +190,7 @@ func (my *Shelf) Run(data *matrix.Matrix) ([]*matrix.Matrix, error) {
 
 	err = rest.FetchData(my.client, href, &records)
 	if err != nil {
-		my.Logger.Error().Stack().Err(err).Str("href", href).Msg("Failed to fetch data")
+		my.Logger.Error().Err(err).Str("href", href).Msg("Failed to fetch data")
 		return nil, err
 	}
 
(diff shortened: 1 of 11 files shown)


cmd/collectors/rest/plugins/shelf/shelf.go
```diff
@@ -316,6 +318,14 @@ func (my *Shelf) Run(data *matrix.Matrix) ([]*matrix.Matrix, error) {
 		return true
 	})
 
+	if len(noSet) > 0 {
+		attributes := make([]string, 0)
+		for k := range noSet {
+			attributes = append(attributes, k)
+		}
```
Hardikl, line 325: maps.Keys(noSet) from golang.org/x/exp/maps?

cmd/collectors/zapi/plugins/qtree/qtree.go
```diff
@@ -214,12 +221,14 @@ func (my *Qtree) Run(data *matrix.Matrix) ([]*matrix.Matrix, error) {
 					for _, label := range my.data.GetExportOptions().GetChildS("instance_keys").GetAllChildContentS() {
 						if value := qtreeInstance.GetLabel(label); value != "" {
 							quotaInstance.SetLabel(label, value)
+							numMetrics++
```
Hardikl, line 224: This is not metric, it just add labels in quota metric from qtree

cmd/collectors/zapi/plugins/qtree/qtree.go
```diff
@@ -231,18 +240,26 @@ func (my *Qtree) Run(data *matrix.Matrix) ([]*matrix.Matrix, error) {
 						if err := m.SetValueString(quotaInstance, value); err != nil {
 							my.Logger.Debug().Msgf("(%s) failed to parse value (%s): %v", attribute, value, err)
 						} else {
+							numMetrics++
```
Hardikl, line 243: This is metric, only this would be required.

Hardikl (approved):

cmd/collectors/rest/plugins/shelf/shelf.go
cgrinds: I thought about that but give the small amount of code needed, it seems better to stick with our own keys instead of pulling in a experimental package.

> Warning: Packages here are experimental and unreliable. Some may one day be promoted to the main repository or other subrepository, or they may be modified arbitrarily or even disappear altogether.

> In short, code in this subrepository is not subject to the Go 1 compatibility promise. (No subrepo is, but the promise is even more likely to be violated by go.exp than the others.)

Hardikl: Got it.

PR #1098, opened 2022-06-14
fix: qtree dashboard fix

rahulguptajss:


conf/rest/9.12.0/qtree.yaml
```diff
@@ -20,7 +22,8 @@ endpoints:
       - ^^qtree           => qtree
       - ^status           => status
       - ^oplock_mode      => oplocks
-
+      - filter:
+          - qtree=!""
```
Hardikl, line 26: It won't be required as endpoint code flow will add labels if matched but it won't do instance creation

Hardikl (approved):

PR #1103, opened 2022-06-16
fix: plugin logger changes

Hardikl:
1. As the plugin code-flow would be invoking few ZAPI/REST calls and add/update the labels in existing metrics based on the response. 
2. Based on the ONTAP version, few ZAPI/REST calls won't work and those labels would be empty as design - NO impact

Changed the plugin logger from Info/Warn to Debug to reduce the log spam for bullet 2 above.


Hardikl:
> I think there will be some genuine errors which may be suppressed in moving all warn to debug. We should move error logging to relevant methods where we log error vs debug.

In other words, below all errors should not be logged as error. 
--> we need to discuss over this because in few cases whole plugin itself is useless when plugin initialization was failed. 

image

PR #1104, opened 2022-06-16
fix: duplicate instance key for SecurityAccount

Hardikl:
In most of the clusters, this field `remote-switch-ipaddress` won't be available as it's based on the remote switch. That's why we haven't encountered in our UTs.

image

**Fix:** added the `remote-switch-ipaddress` as instanceKey
**Detail:** 
- we have restriction in ZAPI workflow that if any key is not exist in response then we ignore/skip that instance
- This logic has been changed, and it will only skip the instance when all the asked keys were not exist in response.
- If one key exist, we would go ahead and parse that instance.


cgrinds:
Verified on `release/22.08.0` a05a0279

PR #1105, opened 2022-06-20
feat: restperf plugins

rahulguptajss:
Pending plugin vscan. It will be developed once templates are ready. Pending from Ontap side.

### DIFF
--- a/cmd/collectors/restperf/plugins/nic/nic.go
+++ b/cmd/collectors/restperf/plugins/nic/nic.go
@@ -1,0 +1,134 @@
+/*
+ *
+Package Description:
+    Some postprocessing on counter data "nic_common"
+      Converts link_speed to numeric MBs
+      Adds custom metrics:
+          - "rc_percent":    receive data utilization percent
+          - "tx_percent":    sent data utilization percent
+          - "util_percent":  max utilization percent
+		  - "nic_state":     0 if port is up, 1 otherwise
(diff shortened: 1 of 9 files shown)


cmd/collectors/restperf/plugins/nic/nic.go
```diff
@@ -0,0 +1,134 @@
...
+func New(p *plugin.AbstractPlugin) plugin.Plugin {
+	return &Nic{AbstractPlugin: p}
+}
+
+func (me *Nic) Run(data *matrix.Matrix) ([]*matrix.Matrix, error) {
+
+	var read, write, rx, tx, util matrix.Metric
+	var err error
+
+	if read = data.GetMetric("receive_bytes"); read == nil {
```
cgrinds, line 37: wasn't the template changed to use `rx_bytes` instead of receive bytes?

cmd/collectors/restperf/plugins/nic/nic.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil, errors.New(errors.ErrNoMetric, "receive_bytes")
+	}
+
+	if write = data.GetMetric("transmit_bytes"); write == nil {
```
cgrinds, line 41: don't think this matches the template

cgrinds (approved):

PR #1127, opened 2022-07-01
fix: dashboard changes for light theme

Hardikl:
Compliance dashboard
Before changes:
image

image

After changes:
image

image

Security dashboard
before changes:
image

image

after changes:
image

image


Hardikl:
Covered
- NTP server number higher as 3 is Compliance
- Datacenter variable is sorted

Hardikl:
> Atleast the first column should have the green sign?

Discussed in call, we are good with only ❌ in first column.

Hardikl:
> Also keeping cross sign before text changes the indentation of records without any sign

Discussed in call, we will go ahead with this indentation for now and based on the customer's feedback, we will revisit this.

cgrinds:
nice job @Hardikl

cgrinds (approved):

PR #1130, opened 2022-07-01
feat: restperf gaps

rahulguptajss:


conf/restperf/9.12.0/system_node.yaml
```diff
@@ -15,19 +16,19 @@ counters:
...
   - disk_data_read
   - disk_data_written
   - hdd_data_read
   - hdd_data_written
   - ssd_data_read
   - ssd_data_written
-  - network_data_received
-  - network_data_sent
-  - fcp_data_received
+  - network_data_received   #net_data_recv in ZapiPerf
```
cgrinds, line 26: per chat, even though REST name is better, we should make these match zapi perf

Hardikl (approved):

PR #1151, opened 2022-07-15
fix: set comment for volume restperf plugin

rahulguptajss:


cgrinds (approved):
does restperf also need a similar fix for https://github.com/NetApp/harvest/blob/de30e117c4673213c7fc41ac44757ddb7944c62b/cmd/collectors/zapiperf/zapiperf.go#L1026

PR #1172, opened 2022-07-27
feat: max plugin

rahulguptajss:


Hardikl (commented):
minor comments

cmd/poller/plugin/README.md
```diff
@@ -111,6 +113,68 @@ The plugin tries to intelligently aggregate metrics based on a few rules:
 - **Weighted Average** - applied if metric has property `average` and suffix `_latency` and if there is a matching `_ops` metric. (This is currently only matching to ZapiPerf metrics, which use the Property field of metrics.)
 - **Ignore** - metrics created by some plugins, such as value_to_num by LabelAgent
 
+# Max
+
+Max creates a new collection of metrics (Matrix) by calculating max of metric values from an existing Matrix for a given label. For example, if the collected metrics are for volumes, you can create max at nodes or svms level.
```
cgrinds, line 118: , you can create a max metric at the node or svm level.

cgrinds: might also be a good idea to show that example or change the text to match one of the existing examples

cmd/poller/plugin/README.md
```diff
@@ -111,6 +113,68 @@ The plugin tries to intelligently aggregate metrics based on a few rules:
...
+
+### Max Rule syntax
+
+simplest case:
+
+```yaml
+plugins:
+  Max:
+    - LABEL
+# will create a new Matrix based consisting of max values on target label LABEL
```
cgrinds, line 128: is this clearer?
> create a new matrix of max values that match the LABEL(s)

cmd/poller/plugin/README.md
```diff
@@ -111,6 +113,68 @@ The plugin tries to intelligently aggregate metrics based on a few rules:
...
+```
+
+If you want to specify which labels should be included in the new instances, you can add those space-seperated after `LABEL`:
+
+```yaml
+    - LABEL LABEL1,LABEL2
+# same, but LABEL1 and LABEL2 will be copied into the new instances
```
cgrinds, line 135: similar to the above example, but ...

cmd/poller/plugin/README.md
```diff
@@ -111,6 +113,68 @@ The plugin tries to intelligently aggregate metrics based on a few rules:
...
+# (default is to only copy LABEL and any global labels (such as cluster and datacenter)
```
cgrinds, line 136: and all global labels

cmd/poller/plugin/README.md
```diff
@@ -111,6 +113,68 @@ The plugin tries to intelligently aggregate metrics based on a few rules:
...
+```
+
+Or include all labels:
+
+```yaml
+    - LABEL ...
+# copy all labels of the original instance
+```
+By default, metrics will be prefixed with `LABEL`. For example if the object of the original Matrix is `volume` (meaning metrics are prefixed with `volume_`) and `LABEL` is `aggr`, then the metric `volume_read_ops` will become `aggr_volume_read_ops`, etc. You can override this by providing the `<>OBJ` using the following syntax:
```
cgrinds, line 145: By default, metrics will be prefixed with `LABEL`. For example if the object of the original Matrix is `volume` (meaning metrics are prefixed with `volume_`) and `LABEL` is `aggr`, then the metric `volume_read_ops` will become `aggr_volume_read_ops`. You can override this using the `<>OBJ` pattern shown below:

cmd/poller/plugin/README.md
```diff
@@ -111,6 +113,68 @@ The plugin tries to intelligently aggregate metrics based on a few rules:
...
+
+```yaml
+    - LABEL<>OBJ
+# use OBJ as the object of the new matrix, e.g. if the original object is "volume" and you
+# want to leave metric names unchanged, use "volume"
+```
+
+Finally, sometimes you only want to generate instances with a specific label value. You can use `<VALUE>` for that (optionally follow by `OBJ`):
```
cgrinds, line 153: > optionally followed by

cmd/poller/plugin/README.md
```diff
@@ -111,6 +113,68 @@ The plugin tries to intelligently aggregate metrics based on a few rules:
...
+    - LABEL<LABELX=`VALUE`>
+# same, but check against "LABELX" (instead of "LABEL")
+```
+
+Examples:
+
+```yaml
+plugins:
+  Max:
+    # will create max of each metrics of the aggr. All metrics will be prefixed with aggr_disk_max. All labels are included in the new instances
```
cgrinds, line 169: > will create max of each aggregate metric

grafana/dashboards/7mode/harvest_dashboard_cluster7.json
```diff
@@ -591,7 +587,7 @@
       ],
       "timeFrom": null,
       "timeShift": null,
-      "title": "Disk busy",
+      "title": "Max Disk Utilization",
```
cgrinds, line 590: is it clearer to say "Max Disk Utilization by Node"?

pkg/matrix/matrix.go
```diff
@@ -184,6 +184,14 @@ func (me *Matrix) RemoveMetric(key string) {
 	delete(me.metrics, key)
 }
 
+func (me *Matrix) RemoveExceptMetric(key string) {
+	for k := range me.metrics {
```
cgrinds, line 188: instead of iterating through the metrics, deleting all but one, can we create a new map with only the one we want to keep. O(1) vs O(n)

cgrinds: this method only touches `metrics`, no need to create a new matrix or if there is this method should do it too. I meant something like this, which is O(1)
```
func (me *Matrix) RemoveExceptMetric(key string) {
	prev, ok := me.metrics[key]
	if !ok {
		return
	}
	me.metrics = make(map[string]Metric)
	me.metrics[key] = prev
}
```

pkg/matrix/matrix.go
```diff
@@ -184,6 +184,15 @@ func (me *Matrix) RemoveMetric(key string) {
 	delete(me.metrics, key)
 }
 
+func (me *Matrix) RemoveExceptMetric(key string) {
+	prev, ok := me.metrics[key]
+	if !ok {
+		return
```
Hardikl, line 190: when key not found then do we need empty matrix or as it is?

Hardikl (approved):

cgrinds (approved):

PR #1178, opened 2022-07-29
fix: fix quota error log

rahulguptajss:


Hardikl (approved):

cgrinds (approved):

cgrinds:
this change reduces log noise - originally, when there were no quotas the code was logging an error. Having no quotas is not an error so move to debug

PR #1184, opened 2022-08-02
feat: lun dashboard should use topk range

cgrinds:


cgrinds:
that's correct, only the six panels in `Top LUN Performance Drilldown`

grafana/dashboards/cmode/harvest_dashboard_lun.json
```diff
@@ -805,7 +805,7 @@
       "targets": [
         {
           "exemplar": false,
-          "expr": "topk($TopResources, lun_avg_write_latency{datacenter=\"$Datacenter\",cluster=\"$Cluster\",svm=~\"$SVM\",volume=~\"$Volume\",lun=~\"$LUN\"})/1000",
```
Hardikl, line 808: why to remove topk from table?

cgrinds: good catch, that should have been reverted

Hardikl (approved):

PR #1194, opened 2022-08-09
docs: release 22.08.0 changelog

rahulguptajss:


ruanruijuan (approved):
"Add filter support for Private Rest CLI" should be under Enhancements.

The EMS collector needs a little bit more description, something like:
Harvest adds ONTAP Event Management System (EMS) collector in this release. It collects ONTAP events and exports to Prometheus database. it also integrates with the Prometheus AlertManager.

cgrinds (approved):

PR #1205, opened 2022-08-16
fix: negative counter values

rahulguptajss:
9:31PM WRN matrix/metric_float64.go:235 > Negative counter detected Poller=u2 collector=ZapiPerf:NFSv3 metric=read_avg_latency minuend=303902176895 subtrahend=303902188591


cgrinds (approved):
thanks, this looks good and better than earlier change

PR #1226, opened 2022-08-23
fix: create unique indexes for quota dashboard

rahulguptajss:


Hardikl (approved):

cmd/collectors/zapi/plugins/qtree/qtree.go
```diff
@@ -140,6 +140,10 @@ func (my *Qtree) Run(data *matrix.Matrix) ([]*matrix.Matrix, error) {
 
 	tag := "initial"
 	quotaIndex := 0
+	var cluster string
+	if data.GetGlobalLabels().Has("cluster") {
```
cgrinds, line 144: minor: one lookup can be done instead of two if we use GetHas

cgrinds (approved):

PR #1229, opened 2022-08-23
fix: Volume dashboard should use IEC bytes

cgrinds:
Per Volume Space Used panel should use IEC bytes
As reported by Falcon667 on Discord


Hardikl:
We need to change the units in table as well to sync with value shown in below graph.
image

Hardikl (approved):

PR #1240, opened 2022-08-26
fix: bin/rest should support verbose and return error when 

### LINKED ISSUE #1230: Possible Issue with the api tool?
If I run rest show data, it works. But If I try to GET param or other attributes of api, it fails with below message. 

Is there a CLI document for this tool?

_bin/harvest rest show models --poller mypoller --api cluster --max-records 2
Using downloaded file /tmp/mypoller_fqdn-swagger.yaml with timestamp 2022-08-23 11:58:42.692790756 -0700 PDT
error unmarshalling swagger file=[ /tmp/mypoller_fqdn-swagger.yaml] err=yaml: unmarshal errors:
  line 1: cannot unmarshal !!str `<!DOCTY...` into map[string]interface {}_

cgrinds:
server responds with !200

Fixes #1230

### DIFF
--- a/cmd/tools/rest/client.go
+++ b/cmd/tools/rest/client.go
@@ -13,6 +13,7 @@ import (
 	"github.com/tidwall/gjson"
 	"io"
 	"net/http"
+	"net/http/httputil"
 	"os"
 	"strings"
 	"time"
@@ -232,7 +233,7 @@ func (c *Client) invoke() ([]byte, error) {
 	return body, nil
 }
 
-func downloadSwagger(poller *conf.Poller, path string, url string) (int64, error) {
+func downloadSwagger(poller *conf.Poller, path string, url string, verbose bool) (int64, error) {
 	var restClient *Client
 
 	out, err := os.Create(path)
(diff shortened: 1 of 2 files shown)


cgrinds:
With this change, when running with a user that does not have the correct role, `bin/rest` prints:
```
error downloading swagger error making request. server response statusCode=[401]
```

With this change, when running with a user that does not have the correct role, `bin/rest --verbose` prints:
```
REQUEST: https://10.193.48.11/docs/api/swagger.yaml BY: harvestnew
GET /docs/api/swagger.yaml HTTP/1.1
Host: 10.193.48.11
User-Agent: Go-http-client/1.1
Authorization: Basic foo
Accept-Encoding: gzip


RESPONSE: 
HTTP/1.1 401 Unauthorized
Content-Length: 381
Content-Security-Policy: default-src 'self'; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; frame-ancestors: 'self'
Content-Type: text/html; charset=iso-8859-1
Date: Sat, 27 Aug 2022 12:38:56 GMT
Server: Apache
Www-Authenticate: Basic realm="ONTAP"
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-Xss-Protection: 1; mode=block

error downloading swagger error making request. server response statusCode=[401]

```

Hardikl (approved):

PR #1246, opened 2022-08-29
fix: ignore all zero performance metrics for nfsv3

### LINKED ISSUE #1245: nfsv3 zero metrics causes spikes in performance metrics
As reported @jmg011 , when all metrics are reported as zero due to a possible ontap bug for nfsv3 then the next poll returns billion ops.

- [x] ZapiPerf Changes
- [x] RestPerf Changes

rahulguptajss:
Fixes #1245

TODO: Rest collector once this PR is merged.


Hardikl (approved):
This approach looks good

PR #1277, opened 2022-09-12
fix: enable dashboard check in CI

rahulguptajss:


Hardikl (dismissed):

cmd/collectors/zapiperf/zapiperf.go
```diff
@@ -579,13 +582,17 @@ func (me *ZapiPerf) PollData() (map[string]*matrix.Matrix, error) {
 			sValues := met.GetValuesFloat64()
 			pass := met.GetPass()
 			for k := range sValues {
-				pass[k] = sValues[k] > 0
+				if me.isCI {
+					pass[k] = sValues[k] > 0
+				} else {
+					pass[k] = sValues[k] >= 0
+				}
```
cgrinds, line 589: The condition is flipped here, I think you meant for this to be
```go
if me.isCI {
    pass[k] = sValues[k] >= 0
} else {
    pass[k] = sValues[k] > 0
}
```
but it might be better to use
```go
pass[k] = isCI || sValues[k] > 0
```

cmd/collectors/restperf/restperf.go
```diff
@@ -782,13 +786,17 @@ func (r *RestPerf) PollData() (map[string]*matrix.Matrix, error) {
 			sValues := m.GetValuesFloat64()
 			pass := m.GetPass()
 			for k := range sValues {
-				pass[k] = sValues[k] > 0
+				if r.perfProp.isCI {
```
cgrinds, line 789: similar comment as in zapiperf.go

cgrinds (approved):

PR #1278, opened 2022-09-12
fix: http sd should publish local IP when exporter is blank

cgrinds:
Reported by Matthias Beck on Discord. Thanks!

### DIFF
--- a/cmd/poller/poller.go
+++ b/cmd/poller/poller.go
@@ -931,10 +931,8 @@ func (p *Poller) publishDetails() {
 			continue
 		}
 		p.hasPromExporter = true
-		if exp.LocalHTTPAddr == "0.0.0.0" {
+		if exp.LocalHTTPAddr == "" || exp.LocalHTTPAddr == "0.0.0.0" {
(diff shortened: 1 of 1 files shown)


cgrinds:
## Config
`harvest.yml` snippet

```yaml
Exporters:
    prometheus1:
        exporter: Prometheus
        port_range: 12990-13000
```

Start poller and admin node. Query service discovery endpoint
```bash
curl -s 'http://localhost:8887/api/v1/sd' | jq .
[
  {
    "targets": [
      ":13002"           <============ this is incomplete and will not work
    ],
    "labels": {
      "__meta_poller": "u2"
    }
  }
]
```

With fix
Start poller and admin node. Query service discovery endpoint
```bash
curl -s 'http://localhost:8887/api/v1/sd' | jq .
[
  {
    "targets": [
      "10.249.132.44:13002"
    ],
    "labels": {
      "__meta_poller": "u2"
    }
  }
]
```

Hardikl (approved):

PR #1304, opened 2022-09-22
feat: harvest should replace topk with topk range in all dashboards Part 3

rahulguptajss:


grafana/dashboards/cmode/harvest_dashboard_nvme_fc.json
```diff
@@ -700,7 +700,7 @@
                 "properties": [
                   {
                     "id": "custom.width",
-                    "value": 184
+                    "value": null
```
cgrinds: was this caused by the export or was this changed?

cgrinds (approved):

PR #1333, opened 2022-10-07
fix: increase default zapi timeout to 30 seconds

rahulguptajss:


Hardikl (approved):

cmd/tools/rest/client.go
```diff
@@ -21,7 +21,7 @@ import (
 
 const (
 	// DefaultTimeout should be > than ONTAP's default REST timeout, which is 15 seconds for GET requests
-	DefaultTimeout = 30
```
cgrinds, line 24: In a few places, we've changed these to durations instead of unit-less ints. Might be worth doing the same here

pkg/api/ontapi/zapi/client.go
```diff
@@ -25,7 +25,7 @@ import (
 
 const (
 	DefaultAPIVersion = "1.3"
-	DefaultTimeout    = 10
```
cgrinds, line 28: same comment

cmd/collectors/rest/rest.go
```diff
@@ -169,15 +182,7 @@ func (r *Rest) getClient(a *collector.AbstractCollector, config *node.Node) (*re
 		r.Logger.Error().Str("poller", opt.Poller).Msg("Address is empty")
 		return nil, errs.New(errs.ErrMissingParam, "addr")
 	}
-
-	clientTimeout := config.GetChildContentS("client_timeout")
-	timeout := rest.DefaultTimeout * time.Second
-	duration, err := time.ParseDuration(clientTimeout)
-	if err == nil {
-		timeout = duration
```
Hardikl, line 177: changes are fine, but could be possible here as well without much modification:
 r.Client.Timeout = timeout

Hardikl (dismissed):

cgrinds (approved):

PR #1344, opened 2022-10-12
feat: matrix changes for perf collector calculations

rahulguptajss:


pkg/matrix/metric_int64.go
```diff
@@ -12,81 +12,78 @@ import (
...
+		if len(m.values) != 0 {
+			clone.values = make(map[string]int64, len(m.values))
+			for key, element := range m.values {
+				clone.values[key] = element
+			}
+		} else {
+			clone.values = make(map[string]int64)
+		}
+	} else {
+		clone.values = make(map[string]int64)
```
Hardikl, line 30: same here

pkg/matrix/metric_int64.go
```diff
@@ -17,14 +17,12 @@ type MetricInt64 struct {
 
 func (m *MetricInt64) Clone(deep bool) Metric {
 	clone := MetricInt64{AbstractMetric: m.AbstractMetric.Clone(deep)}
-	if deep {
+	if deep && len(m.values) != 0 {
 		if len(m.values) != 0 {
```
Hardikl, line 21: missed to remove inner if

Hardikl (approved):

Hardikl (dismissed):
looks cleaner

cgrinds (commented):
minor: All of the LazySet* and LazyGet* methods in lazy.go can be simplified since the metric now includes the instanceKey. 

In essence, three lookups are happening now: one for the metric, one for the instance, and then another for the instance in the metric. The instance lookup in lazy.go can be removed.

PR #1352, opened 2022-10-18
feat: Fabricpool panels - parity with 1.6

Hardikl:


Hardikl:
updated all panels with description and units.
Only pending panel is `Top Volumes with Object Storage Requests` graph, because the variable of type `volName`-`Req type`.

Hardikl:
Added 5 panels as shown below.
image


Me and Rahul discussed regarding the 6th panel where panel would need filter based on `volume-metric`, but due to instances/metrics independent data structure, it's not feasible in backend. Also, I have tried to manage in prometheus query but label_join/label_replace function don't support filter on newly generated label `volume-metric`, and without this it's difficult to implement in panel with topk. 

So, we will submit 5 panels for now and this 6th panel will be targeted later.

PR #1414, opened 2022-11-04
fix: dont export constituents relationships in SM

Hardikl:
Zapi With `expand=false`
image

Zapi With `expand=true` --> got added 16 constituent's relationships here
image

Rest with `expand=false`
image

Rest with `expand=true`
image

### DIFF
--- a/cmd/collectors/zapi/plugins/snapmirror/snapmirror.go
+++ b/cmd/collectors/zapi/plugins/snapmirror/snapmirror.go
@@ -31,6 +31,8 @@ type Peer struct {
 	cluster string
 }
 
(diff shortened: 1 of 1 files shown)


cmd/collectors/zapi/plugins/snapmirror/snapmirror.go
```diff
@@ -69,6 +70,7 @@ func (my *SnapMirror) Run(data *matrix.Matrix) ([]*matrix.Matrix, error) {
 	destUpdCount := 0
 	srcUpdCount := 0
 	limitUpdCount := 0
+	re := regexp.MustCompile(`^(.*)__(\d{4})$`)
```
cgrinds, line 73: better to pull out of the method, so the compile only happens once

PR #1415, opened 2022-11-04
fix: handle LS relationships + handle dashboard

Hardikl:
LS(load sharing) relationships don't have `relationship_id` field in ZAPI/REST response, due to this, we either skip them or wrongly populate in Harvest.

Changes made:
- added instancekey as `relationship_id`+`destination_path`
- make sure ZAPI and REST collector would parse as expected
- make sure no impact in dashboard due to this

With this, We can now monitor LS relationships as well and can show in dashboard if required.

### DIFF
--- a/cmd/collectors/rest/rest.go
+++ b/cmd/collectors/rest/rest.go
@@ -415,13 +415,20 @@ func (r *Rest) HandleResults(result []gjson.Result, prop *prop, isEndPoint bool)
 			continue
 		}
 
-		// extract instance key(s)
-		for _, k := range prop.InstanceKeys {
-			value := instanceData.Get(k)
-			if value.Exists() {
-				instanceKey += value.String()
-			} else {
-				r.Logger.Warn().Str("key", k).Msg("missing key")
+		if len(prop.InstanceKeys) != 0 {
+			// extract instance key(s)
+			for _, k := range prop.InstanceKeys {
+				value := instanceData.Get(k)
+				if value.Exists() {
+					instanceKey += value.String()
+				} else {
+					r.Logger.Warn().Str("key", k).Msg("missing key")
+				}
+			}
+
+			if instanceKey == "" {
+				r.Logger.Trace().Msg("Instance key is empty, skipping")
+				continue
 			}
 		}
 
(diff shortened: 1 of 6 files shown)


cmd/collectors/rest/rest.go
```diff
@@ -421,8 +421,7 @@ func (r *Rest) HandleResults(result []gjson.Result, prop *prop, isEndPoint bool)
 			if value.Exists() {
 				instanceKey += value.String()
 			} else {
-				r.Logger.Warn().Str("key", k).Msg("skip instance, missing key")
-				break
```
Hardikl, line 425: Intention here to make behaviour same as ZAPI.

cgrinds (approved):

cmd/collectors/rest/rest.go
```diff
@@ -421,8 +421,7 @@ func (r *Rest) HandleResults(result []gjson.Result, prop *prop, isEndPoint bool)
...
+				r.Logger.Warn().Str("key", k).Msg("missing key")
```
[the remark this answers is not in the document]
Hardikl: nope.

I we do that, all cluster_scoped one won't created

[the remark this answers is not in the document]
Hardikl: done

cmd/collectors/restperf/restperf.go
```diff
@@ -504,14 +504,20 @@ func (r *RestPerf) PollData() (map[string]*matrix.Matrix, error) {
...
 
-			// extract instance key(s)
-			for _, k := range instanceKeys {
-				value := parseProperties(instanceData, k)
-				if value.Exists() {
-					instanceKey += value.String()
-				} else {
-					r.Logger.Warn().Str("key", k).Msg("skip instance, missing key")
-					break
+			if len(instanceKeys) != 0 {
```
cgrinds, line 507: is this guard clause needed? range handles empty and nil slice

Hardikl: Agreed that range would handle nil and empty. But here, we need to isolate these 2 cases,
1. instancekeys not specified --> `cluster scoped templates` --> here we create instance with empty("") instanceKey and proceed as expected.
2. instancekeys specified but got empty response from ONTAP --> here we log and skip that instance.

PR #1443, opened 2022-11-10
fix: disk dashboard - Variables are not sorted

rahulguptajss:


Hardikl (approved):
side ask: 
`type` var is not used anywhere, can we remove that ?

cgrinds (approved):

PR #1456, opened 2022-11-11
fix: disk type in lower case in Rest

rahulguptajss:


Hardikl (approved):

conf/rest/9.12.0/disk.yaml
```diff
@@ -30,6 +29,7 @@ endpoints:
       - sectors_read            => stats_sectors_read
       - sectors_written         => stats_sectors_written
       - disk_io_kbps_total      => stats_io_kbps
+      - ^type
```
cgrinds, line 32: lovely solution 👍

cgrinds (approved):

PR #1464, opened 2022-11-14
fix: power fix for cold sensors

rahulguptajss:


cmd/collectors/zapi/plugins/sensor/sensor.go
```diff
@@ -134,14 +134,16 @@ func (my *Sensor) calculateEnvironmentMetrics(data *matrix.Matrix) ([]*matrix.Ma
 				}
 
 				if sensorType == "thermal" && !isAmbientMatch {
-					// Exclude temperature sensors that have crit_low=0 and warn_low is missing
-					if !(criticalLowThr == 0.0 && warningLowThr == "") {
-						if value, ok, _ := metric.GetValueFloat64(instance); ok {
+					// Exclude temperature sensors that contains sensor name `Margin` and value < 0
+					value, ok, _ := metric.GetValueFloat64(instance)
+					if !strings.Contains(sensorName, "Margin") && value > 0 {
```
cgrinds, line 139: minor but better to do `value > 0 && !strings.Contains(sensorName, "Margin")`

PR #1488, opened 2022-11-16
fix: net port template fix

rahulguptajss:
operational-speed is string


conf/zapi/cdot/9.8.0/netPort.yaml
```diff
@@ -16,8 +16,8 @@ counters:
     - ^vlan-id                                 => vlan_id
     - ^vlan-node                               => vlan_node
     - ^vlan-port                               => vlan_port
+    - ^^operational-speed                      => op_speed
```
Hardikl, line 19: do we need as `^^`?

cgrinds (approved):

PR #1501, opened 2022-11-17
fix: remove queue_full counter from namespace template

rahulguptajss:


cgrinds:
This counter was removed because it does not exist in namespace.
```
curl -k -n 'https://10.195.15.41/api/cluster/counter/tables/namespace?return_records=true&fields=*' | rg queue_full
```

PR #1547, opened 2022-11-25
fix: handle alignment in svm panel

Hardikl:


Hardikl:
Handled 1 and 2 points mentioned above.

For handling 3, it will be many hidden vars same as volume dashboard, Working on it.

Hardikl:
Empty rows are handled in recent commit. 
legends are already there, may be dashboard wasn't updated. 
Added `SVM updated` dashboard in both the machines, shared via chat.

Issue #1558, opened 2022-11-28
Volume metrics in cDOT dashboard are missing legend


rahulguptajss:
![image](https://user-images.githubusercontent.com/25551691/204366366-fe4fbce0-2dec-48fd-adf3-8f1f05858f65.png)

Hardikl:
will be handled in #1576

PR #1579, opened 2022-12-01
feat: added dashboard tests for legends details

cgrinds:
Fixes: #1558

### DIFF
--- a/cmd/tools/grafana/dashboard_test.go
+++ b/cmd/tools/grafana/dashboard_test.go
@@ -653,59 +653,60 @@ func TestLegends(t *testing.T) {
 func checkLegends(t *testing.T, path string, data []byte) {
 	// collect all legends
 	dashPath := shortPath(path)
-	gjson.GetBytes(data, "panels").ForEach(func(key, value gjson.Result) bool {
+
+	visitAllPanels(data, func(path string, key, value gjson.Result) {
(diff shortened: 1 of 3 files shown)


cmd/tools/grafana/dashboard_test.go
```diff
@@ -653,59 +653,60 @@ func TestLegends(t *testing.T) {
...
 		legendPlacementData := legendData.Get("placement").String()
 		title := value.Get("title").String()
-		if calcsData := legendData.Get("calcs"); calcsData.Exists() {
-			var calcsSlice []string
-			calcsData.ForEach(func(key, val gjson.Result) bool {
-				calcsSlice = append(calcsSlice, val.String())
-				return true
-			})
-			checkCalcs(t, calcsSlice, expectedCals, dashPath, title)
+		calcsData := legendData.Get("calcs").Array()
```
Hardikl, line 675: few panels don't have `calc` object

Hardikl (approved):

cmd/tools/grafana/dashboard_test.go
Hardikl: tested and `test` passes with this.

cgrinds: right, it's fine when `calcs` is missing - we get an empty array in that case

PR #1598, opened 2022-12-07
fix: histogram issues

rahulguptajss:
Fixed below issues

1: Noticed that description may have capital `H` in RestPerf which breaks histogram logic.
2: Newly introduced `else block` in metric calculations captures `.bucket` metric as `skipped` which increases skip count. `.bucket` metric needs no calculation hence skipped from calculation.
3: Added `key` in logs of metric calculation for array/histogram metrics.
4: Added a check if display label passed is empty then use key as metric name.

### DIFF
--- a/cmd/collectors/restperf/restperf.go
+++ b/cmd/collectors/restperf/restperf.go
@@ -640,6 +640,7 @@ func (r *RestPerf) PollData() (map[string]*matrix.Matrix, error) {
 			for name, metric := range r.Prop.Metrics {
 				f := parseMetricResponse(instanceData, name)
 				if f.value != "" {
+					description := strings.ToLower(r.perfProp.counterInfo[name].description)
 					// special case for workload_detail
 					if isWorkloadDetailObject(r.Prop.Query) {
 						if name == "wait_time" || name == "service_time" {
@@ -680,7 +681,7 @@ func (r *RestPerf) PollData() (map[string]*matrix.Matrix, error) {
 							// ONTAP does not have a `type` for histogram. Harvest tests the `desc` field to determine
 							// if a counter is a histogram
 							isHistogram = false
-							if len(labels) > 0 && strings.Contains(strings.ToLower(r.perfProp.counterInfo[name].description), "histogram") {
+							if len(labels) > 0 && strings.Contains(description, "histogram") {
 								key := name + ".bucket"
 								histogramMetric = curMat.GetMetric(key)
 								if histogramMetric != nil {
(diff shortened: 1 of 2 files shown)


cmd/collectors/zapiperf/zapiperf.go
```diff
@@ -557,13 +557,13 @@ func (z *ZapiPerf) PollData() (map[string]*matrix.Matrix, error) {
 	orderedKeys := make([]string, 0, len(orderedMetrics))
 
 	for key, metric := range curMat.GetMetrics() {
-		if metric.GetComment() == "" { // does not require base counter
+		if metric.GetComment() == "" && !strings.Contains(key, ".bucket") { // does not require base counter
```
cgrinds, line 560: minor: but it's likely faster and bit clearer to use metric.Buckets() == nil

cmd/collectors/zapiperf/zapiperf.go
```diff
@@ -557,13 +557,13 @@ func (z *ZapiPerf) PollData() (map[string]*matrix.Matrix, error) {
...
 			orderedMetrics = append(orderedMetrics, metric)
 			orderedKeys = append(orderedKeys, key)
 		}
 	}
 	for key, metric := range curMat.GetMetrics() {
-		if metric.GetComment() != "" { // requires base counter
+		if metric.GetComment() != "" && !strings.Contains(key, ".bucket") { // requires base counter
```
cgrinds, line 566: minor: but it's likely faster and bit clearer to use metric.Buckets() == nil

cmd/collectors/zapiperf/zapiperf.go
```diff
@@ -1168,7 +1168,7 @@ func (z *ZapiPerf) addCounter(counter *node.Node, name, display string, enabled
 		// ONTAP does not have a `type` for histogram. Harvest tests the `desc` field to determine
 		// if a counter is a histogram
 		isHistogram = false
-		if len(labels) > 0 && strings.Contains(description, "histogram") {
+		if len(labels) > 0 && strings.Contains(strings.ToLower(description), "histogram") {
```
cgrinds, line 1171: can we `toLower` once, when we set the description instead of on every iteration?

cmd/collectors/restperf/restperf.go
```diff
@@ -680,7 +680,7 @@ func (r *RestPerf) PollData() (map[string]*matrix.Matrix, error) {
 							// ONTAP does not have a `type` for histogram. Harvest tests the `desc` field to determine
 							// if a counter is a histogram
 							isHistogram = false
-							if len(labels) > 0 && strings.Contains(r.perfProp.counterInfo[name].description, "histogram") {
+							if len(labels) > 0 && strings.Contains(strings.ToLower(r.perfProp.counterInfo[name].description), "histogram") {
```
cgrinds, line 683: can we toLower once, when we set the description instead of on every iteration?

cmd/collectors/restperf/restperf.go
```diff
@@ -828,7 +828,7 @@ func (r *RestPerf) PollData() (map[string]*matrix.Matrix, error) {
 	orderedDenominatorKeys := make([]string, 0, len(orderedDenominatorMetrics))
 
 	for key, metric := range curMat.GetMetrics() {
-		if metric.GetName() != "timestamp" {
+		if metric.GetName() != "timestamp" && !strings.Contains(key, ".bucket") {
```
cgrinds, line 831: i made this comment in zapiperf and I think the same applies here 
> minor: but it's likely faster and bit clearer to use metric.Buckets() == nil

PR #1621, opened 2022-12-15
feat: add failed sensors to power dashboard

rahulguptajss:
1: Added a new panel in power dashboard which shows **NOT** normal sensors

image

2: Changed template to add critical_low and critical_high as labels just like warning_low/warning_high. We don't use them in dashboards.

3. Removed color encoding from max/min/avg temperature in grafana.

### DIFF
--- a/grafana/dashboards/cmode/power.json
+++ b/grafana/dashboards/cmode/power.json
@@ -1057,7 +1057,7 @@
           "type": "table"
         }
       ],
-      "title": "Sensors Problems",
+      "title": "Sensor Problems",
       "type": "row"
     },
     {


grafana/dashboards/cmode/power.json
```diff
@@ -976,628 +944,658 @@
...
-          {
-            "matcher": {
-              "id": "byName",
-              "options": "Avg Temp (C)"
+            }
+          ],
+          "type": "table"
+        }
+      ],
+      "title": "Sensors Problems",
```
cgrinds, line 1060: maybe Sensor Problems or Failed Sensors

cgrinds (approved):

PR #1631, opened 2022-12-16
feat: include storagegrid dashboard in docker compose

cgrinds:
Fixes: #1623

### DIFF
--- a/grafana/dashboards/storageGrid.yml
+++ b/grafana/dashboards/storageGrid.yml
@@ -1,12 +1,12 @@
 apiVersion: 1
 
 providers:
-  - name: 'Harvest 7mode'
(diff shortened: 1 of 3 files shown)


grafana/dashboards/storageGrid.yml
```diff
@@ -0,0 +1,12 @@
+apiVersion: 1
+
+providers:
+  - name: 'Harvest cDOT'
```
Hardikl, line 4: Harvest StorageGrid?

cgrinds: Doesn't seem like this name matters? With the changes above, the dashboards are imported like so: 
image

PR #1642, opened 2022-12-23
feat: added 3 panels supported with relationshipId data link

Hardikl:
Changes:
- Added hidden SnapMirrorID var
- Added 3 panels: `Transfer Duration`, `Lag Time` and `Transfer Data` with SnapMirrorID filter support
- Added ID column in `Last Transfer` table with data link - which passes Datacenter, SourceCluster and SnapMirrorID.

SnapMirror Dashboard:
image

When Clicked on ID link:
image

Open issue:
Grafana dynamic link seems to be changed always when we do re-import, so again we need to update the data link of ID column accordingly. Need to check if any feasible solution for this.


Hardikl:
As we passing the empty ("") uid in all the dashboards, so Grafana would provide unique uid for them at each import/re-import.
Due to this, data link of any panel or dashboard won't be same across import. Hence, we can't support data link.

Changing approach to add Id variable along with the Id column in table and based on that var, newly added 3 panels will be filtered.

image

image

image

Hardikl:
Added description for 4 panels as: 
`Must select Source Volume/Destination Volume for detail of single relationship`

Full dashboard:
image


If select source/destination volume:
image

PR #1658, opened 2023-01-12
fix: exclude node/svm vols and include data svms

Hardikl:
Naming conventions
volume
- svm_root volume name: xxx_root
- node_root volume name: vol0

SVM
- system_svm: cluster
- node_svm: node name
- admin_svm: cluster name

**Before applying on template level filter**

Volume:
svm_root: 16
image

node_root:2
image

total volume:527
image

SVM:
data svm:17
image

node,system,admin svm: 4
image

total svm: 21
image

**After applying the template level filter**
Volume:
total volume: 509
image

SVM:
total svm: 17
image

### DIFF
--- a/conf/zapi/cdot/9.8.0/svm.yaml
+++ b/conf/zapi/cdot/9.8.0/svm.yaml
@@ -8,6 +8,7 @@ counters:
     - ^^uuid                                  => svm_uuid
     - ^vserver-name                           => svm
     - ^state                                  => state
+    - ^vserver-type                           => type
     - ^anti-ransomware-default-volume-state   => anti_ransomware_state
     - name-server-switch:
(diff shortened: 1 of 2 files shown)


Hardikl:
For REST,

- Volume public api provide all volumes but node, We only need to exclude root svm vols in template.
- SVM private api provide all svms but system, We only need to include data svms in template.

Hardikl:
> I think we should export root vols/svm where applicable and only hide them from dashboards?

Yes, We should. 
The thought process behind this template filter is, If customer want to exclude/include root objs, they can do via template change which would be easier than query changes in dashboard.

Hardikl:
image

cgrinds (approved):
Nice job @Hardikl

PR #1677, opened 2023-01-30
feat: add StorageGRID overview dashboard

cgrinds:
feat: add StorageGRID Prometheus collector


cmd/tools/grafana/grafana_test.go
```diff
@@ -10,16 +10,25 @@ import (
...
+		{version: "7.1.0", want: true},
+		{version: "7.5.5", want: true},
+	}
+	for _, check := range checks {
+		got := checkVersion(check.version)
+		if got != check.want {
+			t.Errorf("Expected %t but got %t for input %s", check.want, got, check.version)
 		}
 	}
+	t.Log("") // required so the test is not marked as terminated
```
Hardikl, line 31: is this needed for other places as well ?

Hardikl (approved):

ewilts:
Can we also get graphs by storage pools?  Although I personally have only 1 pool per grid, I would expect that other customers may have multiple pools.

PR #1679, opened 2023-02-01
fix: backward compatibility for qtree template

Hardikl:
In 22.05,

qtree_labels:
image

qtree_disk_llimit:
image

No quota_xxx metrics exist.

In 23.02, without any changes,
qtree_labels:
image

qtree_disk_limit:
image

quota_disk_limit:
image

In 23.02, with changes with qtree.yaml,
qtree_labels:
image

qtree_disk_limit:
image

quota_disk_limit:
image

In 23.02, when changed from qtree.yaml to qtree.22.05.yaml in default.yaml
qtree_labels:
image

qtree_disk_limit:
image

quota_disk_limit:
image

Points:
- labels of `qtree_labels` would be same in both the template.
- no of records of `qtree_labels` would be differ in both template --> reason: latest template excludes empty qtree names.
- no of records of quota metrics would be differ in both template --> reason: latest template excludes empty qtree names.

Current Issue:
- latest template would expect all type of quota when no quota-type filter applied in template, whereas old template would expect only tree quotas.


Hardikl:
Added minor change in older template to fetch only `tree` quotas as Harvest 22.05 only gives tree quotas.

**Reason**: We have only one `Qtree` plugin which also contains all the latest changes. It would be easier to get same metric behaviour from the template change than plugin changes. 

Metrics when changed to qtree.22.05.yaml in default.yaml.
qtree_labels:
image

qtree_disk_limit:
image

quota_disk_limit:
image


Points:
- labels of qtree_labels would be same in both the template.
- 3 extra labels[index, type, unit] would be present in qtree_xxx metric in 22.05 template 
- no of records of qtree_labels, qtree_xxx would be same in both the template.
- qtree.22.05.yaml gives same behaviour as Harevst 22.05
- qtree.yaml gives current behaviour

Hardikl:
Case 1: **Same behaviour as Harvest 22.05**
with hidden key-val pair
qtree_labels:
image

qtree_disk_limit:
image

quota_xxxx metrics not exist.



Case 2: **Current Harvest behaviour** 
without hidden key-val pair:
qtree_labels:
image

qtree_disk_limit:
image

quota_disk_limit:
image

Hardikl:
Change update in PR:
with hidden flag --> both qtree and quota metrics will be visible with new labels as well as older labels.
without hidden flag --> both qtree and quota metrics will be visible with new labels only.

cmd/collectors/zapi/plugins/qtree/qtree.go
```diff
@@ -61,10 +62,28 @@ func (my *Qtree) Init() error {
...
-	objects := my.Params.GetChildS("objects")
-	if objects == nil {
-		return errs.New(errs.ErrMissingParam, "objects")
+	if my.Params.HasChildS("historicalLabels") {
+		exportOptions := node.NewS("export_options")
+		instanceKeys := exportOptions.NewChildS("instance_keys", "")
+
+		// apply all instance keys, instance labels from parent (qtree.yaml) to all quota metrics
+		//parent instancekeys would be added in plugin metrics
+		for _, parentKeys := range my.ParentParams.GetChildS("export_options").GetChildS("instance_keys").GetAllChildContentS() {
```
cgrinds, line 73: are we sure that each of these nodes will be present, otherwise we will panic on nil

cmd/collectors/zapi/plugins/qtree/qtree.go
```diff
@@ -61,10 +62,28 @@ func (my *Qtree) Init() error {
...
+			instanceKeys.NewChildS("", parentKeys)
+		}
+		// parent instacelabels would be added in plugin metrics
+		for _, parentLabels := range my.ParentParams.GetChildS("export_options").GetChildS("instance_labels").GetAllChildContentS() {
```
cgrinds, line 77: similar node comment here

Hardikl: same as above

cmd/collectors/zapi/plugins/qtree/qtree.go
Hardikl: It's export section from qtree template, it should always be there to define which keys/labels would be exported for qtree.
 
image

Hardikl: added

cgrinds (approved):

PR #1688, opened 2023-02-03
fix: variable DS_PROMETHEUS should exist in dashboard

cgrinds:
test: add unit test to ensure variable exists

### DIFF
--- a/harvest.yml
+++ b/harvest.yml
@@ -61,7 +61,7 @@ Pollers:
   # no need to add collectors, since they're already defined in Defaults
 
   # example with certificate authentication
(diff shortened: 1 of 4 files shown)


harvest.yml
```diff
@@ -70,4 +70,13 @@ Pollers:
...
+  #     - prometheus1
+
+  # StorageGRID example
+  #  grid-01:
+  #    datacenter: DC-01
+  #    addr: 10.0.0.3
+  #    username: myuser
+  #    password: mypass
+  #    collectors:
+  #      - StorageGrid
```
Hardikl, line 82: do we want to mention example of another prometheus which has storagegrid metrics exported.

PR #1698, opened 2023-02-08
feat: calculate power per TB and Watt

rahulguptajss:


cgrinds (approved):
nice job tracking this down and implementing @rahulguptajss

Hardikl (approved):

PR #1706, opened 2023-02-09
feat: support node aggregation for flexgroup also

Hardikl:
Node aggregation for volume perf metrics are used in these panels:
1. Node dashbord: 
Backend Drilldown panels: -> Values are matching with 1.6

2. Cdot dashboard:
Used avg of node_volume_avg_latency.  

3. Cluster dashboard:
Highlight panels: --> values are matching with 1.6

Side changes:
- one query `WAFL Write Cleaning` was hidden in 2.0, whereas 1.6 showing the same.
- changed few units such that same metric would have same units for all dashboard.


cgrinds (approved):
nice catch and fix @Hardikl

PR #1731, opened 2023-02-14
fix: add error log when collector fails due to connection issues

rahulguptajss:


Hardikl (approved):

cgrinds:
Can you make sure this is not handled lower in the stack? I thought it was, and that's why we were returning a Harvest error.

PR #1733, opened 2023-02-14
feat: Topk panels should use topResources var in their titles

cgrinds:


grafana/dashboards/cmode/volume.json
```diff
@@ -164,7 +164,7 @@
       ],
       "timeFrom": null,
       "timeShift": null,
-      "title": "Volume Average Latency",
+      "title": "Top $TopResources Volumes by Average Latency",
```
Hardikl, line 167: it's stat panel which show one value of avg of topN, title may not reflect that.

grafana/dashboards/cmode/volume.json
```diff
@@ -222,7 +222,7 @@
       ],
       "timeFrom": null,
       "timeShift": null,
-      "title": "Volume Total Throughput",
+      "title": "Top $TopResources Volumes by Total Throughput",
```
Hardikl, line 225: may be Top $TopResources volumes Total Throughput?

Hardikl (approved):

PR #1735, opened 2023-02-15
fix: topk suppport in 3 panels in snapmirror dashboard

Hardikl:
image

image

### DIFF
--- a/grafana/dashboards/cmode/snapmirror.json
+++ b/grafana/dashboards/cmode/snapmirror.json
@@ -882,7 +882,7 @@
       "pluginVersion": "8.1.8",
       "targets": [
         {
-          "expr": "avg (snapmirror_last_transfer_duration{source_cluster=~\"$SourceCluster\",source_volume=~\"$SourceVolume\",destination_volume=~\"$DestinationVolume\"}) by (relationship_id)",
+          "expr": "snapmirror_last_transfer_duration{source_cluster=~\"$SourceCluster\",source_volume=~\"$SourceVolume\",destination_volume=~\"$DestinationVolume\"}",
           "format": "table",
           "instant": true,
           "interval": "",
@@ -890,7 +890,7 @@
           "refId": "C"
         },
         {
-          "expr": "avg(snapmirror_last_transfer_size{source_cluster=~\"$SourceCluster\",source_volume=~\"$SourceVolume\",destination_volume=~\"$DestinationVolume\"}) by (relationship_id)",
+          "expr": "snapmirror_last_transfer_size{source_cluster=~\"$SourceCluster\",source_volume=~\"$SourceVolume\",destination_volume=~\"$DestinationVolume\",relationship_id=~\"$TopTransferDataById\"}",
           "format": "table",
           "instant": true,
           "interval": "",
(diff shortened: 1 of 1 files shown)


grafana/dashboards/cmode/snapmirror.json
```diff
@@ -882,15 +882,15 @@
       "pluginVersion": "8.1.8",
       "targets": [
         {
-          "expr": "avg (snapmirror_last_transfer_duration{source_cluster=~\"$SourceCluster\",source_volume=~\"$SourceVolume\",destination_volume=~\"$DestinationVolume\"}) by (relationship_id)",
```
Hardikl, line 885: avg by relationship_id doesn't do any work as relationship_id is unique

grafana/dashboards/cmode/snapmirror.json
```diff
@@ -882,15 +882,15 @@
...
+          "expr": "snapmirror_last_transfer_duration{source_cluster=~\"$SourceCluster\",source_volume=~\"$SourceVolume\",destination_volume=~\"$DestinationVolume\"}",
           "format": "table",
           "instant": true,
           "interval": "",
           "legendFormat": "",
           "refId": "C"
         },
         {
-          "expr": "avg(snapmirror_last_transfer_size{source_cluster=~\"$SourceCluster\",source_volume=~\"$SourceVolume\",destination_volume=~\"$DestinationVolume\"}) by (relationship_id)",
+          "expr": "snapmirror_last_transfer_size{source_cluster=~\"$SourceCluster\",source_volume=~\"$SourceVolume\",destination_volume=~\"$DestinationVolume\",relationship_id=~\"$TopTransferDataById\"}",
```
cgrinds, line 893: it feels a bit surprising that the `Last Transfers` table is filtered by topk size. We should probably update the title of this panel to make that clear. Not sure folks will understand that the number of rows shown is a function of topk

cgrinds:
Not sure if this was introduced by your change, but looks like a column is missing?
image

grafana/dashboards/cmode/snapmirror.json
Hardikl: Changed to `Top $TopResources Last Transfers by Transfer Data Size`

Hardikl:
> Not sure if this was introduced by your change, but looks like a column is missing? image

it was there earlier, and not related to this table, so removed them.

Hardikl:
image

grafana/dashboards/cmode/snapmirror.json
Hardikl: added filter for all columns.

For `all` option in topk,
Currently 500 is max value in topk. May be we provide highest value as some ~K, which gives all records.

Hardikl: As `topk(100, snapmirror_labels)` function only allow k number as arg, I would be thinking this way:  
image

Hardikl: Tried this, But topk query would fail with this because not able to parse `|`.

image

image

image

Hardikl: Missed the custom all value.... 
made the changes for `All`.

PR #1739, opened 2023-02-15
fix: published metrics can not contain dashes

### LINKED ISSUE #1738: Revisit counter dash to underscore replacement logic
As raised by jfong on [Discord](https://discord.com/channels/855068651522490400/1075465999781986364/1075465999781986364)

This logic happens [here](https://github.com/NetApp/harvest/blob/1be3ef44e7f65506e76f74f81f6d478b522f648c/pkg/util/util.go#L251)

Replacement of dash to underscore happens automatically when you don't use the rename  `=>` marker, and yes, that behavior is used extensively in zapi templates. Perhaps the idea is if you rename, you get the final call to decide on the name (even if it's wrong). That may be questionable 🙂

As [noted](https://netapp.github.io/harvest/23.02/resources/templates-and-metrics/#prometheus-wire-format) Prometheus does not permit dashes `-` in labels.

cgrinds:
Fixes #1738

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -239,9 +239,12 @@ func Intersection(a []string, b []string) ([]string, []string) {
 	return matches, misses
 }
 
+var displayReplacer = strings.NewReplacer("-", "_", ".", "_")
+
 // ParseMetric parses display name and type of field and metric type from the raw name of the metric as defined in (sub)template.
 // Users can rename a metric with "=>" (e.g. some_long_metric_name => short).
 // Trailing "^" characters are ignored/cleaned as they have special meaning in some collectors.
+// Dashes (-) and periods (.) will be replaced with underscores in display names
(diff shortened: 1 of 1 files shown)


pkg/util/util.go
```diff
@@ -255,9 +258,9 @@ func ParseMetric(rawName string) (string, string, string, string) {
 		name, metricType = ParseMetricType(name)
 	} else {
 		name = rawName
-		display = strings.ReplaceAll(rawName, ".", "_")
-		display = strings.ReplaceAll(display, "-", "_")
+		display = rawName
 	}
+	display = displayReplacer.Replace(display)
```
Hardikl, line 263: This also need to be changed: https://github.com/NetApp/harvest/blob/047dd568d14987fbbe75daa6fcdb8d419684e465/cmd/collectors/zapiperf/zapiperf.go#L857 and line 1058

RestPerf as well.

PR #1741, opened 2023-02-16
fix: changed filter from not nil to greater than 0

Hardikl:
Tested in 3 machines.


cgrinds (approved):

Hardikl:
Imported the json with changes to .74 and .86 system name as `Snapmirror Updated` dashboard for testing

PR #1777, opened 2023-02-24
feat: dependabot should bump dependencies

cgrinds:


cgrinds:
Let's start with daily and increase if needed

PR #1781, opened 2023-02-27
fix: aggregating latency metrics returns NaN when base counter is 0

rahulguptajss:


Hardikl (approved):

cmd/poller/plugin/aggregator/aggregator.go
```diff
@@ -274,8 +276,15 @@ func (a *Aggregator) Run(data *matrix.Matrix) ([]*matrix.Matrix, error) {
 					continue
 				}
 
-				if err = metric.SetValueFloat64(instance, v/count); err != nil {
-					a.Logger.Error().Stack().Err(err).Msgf("set value [%s] [%s]:", mn, key)
+				// if no ops happened
```
cgrinds, line 279: minor: maybe slightly clearer?
```go
// if no ops happened
if count == 0 {
    err = metric.SetValueFloat64(instance, 0)
} else {
    err = metric.SetValueFloat64(instance, v/count)
}
if err != nil {
    a.Logger.Error().Err(err).Str("mn", mn).Str("key", key).Msg("set value")
}
```

cgrinds (approved):

Issue #1784, opened 2023-02-27
Ensure qtree REST template metrics matches ZAPI


cgrinds:
See also #1678

Hardikl:
This is the way Harvest would publish qtree metrics/labels the same as Harvest 22.05 did, you can follow these steps.

1. Create a `conf/rest/custom.yaml` file, if it doesn't already exist, and add the following lines:

```yaml
objects:
 Qtree: custom_qtree.yaml
```

2. Create a new object template `conf/rest/9.12.0/custom_qtree.yaml` with the following lines.

```yaml
name:                       Qtree
query:                      api/storage/qtrees
object:                     qtree

client_timeout: 2m

counters:
  - ^^svm.name            => svm
  - ^^volume.name         => volume
  - ^^name                => qtree
  - ^export_policy.name   => export_policy
  - ^security_style       => security_style
  - id                    => id


endpoints:
  - query: api/private/cli/qtree
    counters:
      - ^^vserver         => svm
      - ^^volume          => volume
      - ^^qtree           => qtree
      - ^status           => status
      - ^oplock_mode      => oplocks

plugins:
  - LabelAgent:
      exclude_equals:
        - qtree ``
      replace:
        - oplocks oplocks `enable` `enabled`
  - Qtree:
      quotaType:
        - tree
#        - user
#        - group
      historicalLabels: true

export_options:
  instance_keys:
    - qtree
    - volume
    - svm
  instance_labels:
    - export_policy
    - security_style
    - status
    - oplocks
```

3. When you restart your pollers, Harvest will ignore the out-of-the-box `qtree.yaml` template and use the `custom_qtree.yaml` one.

The `historicalLabels` key causes Harvest to publish qtree metrics and labels with both the 22.05 and 23.02 shapes. More details on merging can be found [here](https://netapp.github.io/harvest/23.02/configure-templates/#how-to-extend-a-restrestperfems-collectors-existing-object-template).

Hardikl:
Tested on 23.05

Behaviour without flag: shape and count is same in both zapi/rest collector.
```
curl -s http://localhost:13006/metrics | grep qtree_labels
qtree_labels{cluster="A250-41-42-43",datacenter="zapi",export_policy="default",oplocks="enabled",qtree="harvestqtree",security_style="unix",status="normal",svm="astra_ci_vc_esxi_24_75_data",volume="astra_ci_vc_esxi_24_75_data_root"} 1.0
[...]

PR #1788, opened 2023-03-01
fix: backward compatibility for qtree metrics in rest

Hardikl:
Tested with REST collector, 
metrics behaviour in Harvest 22.05 

**qtree_labels:**
image

**qtree_disk_limit:**
image

**quota_disk_limit:**
image

metrics behaviour in latest Harvest
Case1: **Same behaviour as Harvest 22.05**
with hidden key-val pair

**qtree_labels:**
image

**qtree_disk_limit:**
image

**quota_disk_limit:**
image

Case2: **Current Harvest behaviour**
without hidden key-val pair:

**qtree_labels:**
image

**qtree_disk_limit:**
image

**quota_disk_limit:**
image

### DIFF
--- a/conf/rest/9.12.0/qtree.yaml
+++ b/conf/rest/9.12.0/qtree.yaml
@@ -11,8 +11,6 @@ counters:
   - ^export_policy.name   => export_policy
   - ^security_style       => security_style
   - id                    => id
-  - filter:
-      - name=!""
 
 
 endpoints:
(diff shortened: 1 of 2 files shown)


conf/rest/9.12.0/qtree.yaml
```diff
@@ -11,8 +11,6 @@ counters:
   - ^export_policy.name   => export_policy
   - ^security_style       => security_style
   - id                    => id
-  - filter:
-      - name=!""
```
Hardikl, line 15: As 22.05 behaviour would need all qtrees, I have removed this filter and later filtered with `exclude_equals`

Hardikl: Yes, can be possible.

As the REST don't support merge of custom template, We can ask them to create this custom rest template as mentioned here https://github.com/NetApp/harvest/issues/1784, I will update the PR and the steps accordingly so no changes would be required in current template.

Hardikl: Updated the PR, we don't need to change the steps mentioned in #1784.

PR #1840, opened 2023-03-21
feat: Add routes to network dashboard

rahulguptajss:
image

### DIFF
--- a/grafana/dashboards/cmode/network.json
+++ b/grafana/dashboards/cmode/network.json
@@ -36,6 +36,12 @@
     },
     {
       "type": "panel",
+      "id": "text",
+      "name": "Text",
+      "version": ""
+    },
+    {
+      "type": "panel",
       "id": "timeseries",
       "name": "Time series",
       "version": ""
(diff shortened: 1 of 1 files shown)


cmd/collectors/rest/plugins/netroute/netroute.go
```diff
@@ -49,6 +50,7 @@ func (n *NetRoute) Init() error {
 			iLabels.NewChildS("", label)
 		}
 		iKeys.NewChildS("", "index")
+		iKeys.NewChildS("", "route_uuid")
```
Hardikl, line 53: where does the `uuid` and `route_uuid` would be used in table?

Hardikl: ohk, it's in plugin generated metrics. Got it.

Hardikl (approved):

cgrinds:
Since the Routes panel in the dashboard only works with the RestCollector, we should note that in the Routes row

cgrinds (approved):

PR #1847, opened 2023-03-22
feat: Harvest should collect adaptive QOS policy groups

cgrinds:
Thanks to faguayot for reporting


cmd/collectors/zapi/plugins/qospolicyfixed/qospolicyfixed_test.go
```diff
@@ -11,6 +11,8 @@ func Test_zapiXputToRest(t *testing.T) {
 		want  MaxXput
 		isErr bool
 	}{
+		{zapi: "6144IOPS/TB", want: MaxXput{IOPS: "6144", Mbps: "0"}},
```
Hardikl, line 14: Just add comment as this is a test case of adaptive qos policy

Hardikl (approved):

PR #1881, opened 2023-03-31
feat: cluster health dashboard

rahulguptajss:


cmd/collectors/rest/plugins/health/health.go
```diff
@@ -0,0 +1,189 @@
...
+
+	if err = v.client.Init(5); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (v *Health) initAllMatrix() error {
+	v.data = make(map[string]*matrix.Matrix)
```
Hardikl, line 66: we do reset map in next function, so not needed here

cmd/collectors/rest/plugins/health/health.go
```diff
@@ -0,0 +1,189 @@
...
+	if err := v.initMatrix(diskHealthMatrix); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (v *Health) initMatrix(name string) error {
+	v.data = make(map[string]*matrix.Matrix)
+	v.data[name] = matrix.New(v.Parent+name, name, name)
```
Hardikl, line 75: may be directly use `diskHealthMatrix` here, so `initMatrix` only needed and we don't need `initAllMatrix`.

grafana/dashboards/cmode/health.json
```diff
@@ -0,0 +1,620 @@
...
+          "id": 232,
+          "options": {
+            "showHeader": true
+          },
+          "pluginVersion": "8.1.8",
+          "targets": [
+            {
+              "exemplar": true,
+              "expr": "health_disk_alerts{datacenter=~\"$Datacenter\",cluster=~\"$Cluster\"} * on(disk,cluster,datacenter) group_left(serial_number,model,shelf) disk_labels{datacenter=~\"$Datacenter\",cluster=~\"$Cluster\"}",
+              "format": "table",
```
Hardikl, line 363: minor: If 2-3 fields from join then either we can do in rest call itself to avoid this join. 
as this is disk object we won't gain much in both the ways.

cmd/collectors/rest/plugins/health/health.go
```diff
@@ -0,0 +1,189 @@
...
+func (v *Health) collectDiskAlerts() {
+	var (
+		instance *matrix.Instance
+	)
+	if records, err := v.getDisks(); err != nil {
+		if errs.IsRestErr(err, errs.APINotFound) {
+			v.Logger.Debug().Err(err).Msg("API not found")
+		} else {
+			v.Logger.Error().Err(err).Msg("Failed to collect analytic data")
+		}
```
cgrinds, line 143: maybe return when there is an err and unwrap the else block

cmd/collectors/rest/plugins/health/health.go
```diff
@@ -0,0 +1,189 @@
...
+)
+
+type AlertSeverity string
+
+const (
+	errr    AlertSeverity = "error"
+	warning AlertSeverity = "warning"
+)
+
+const diskHealthMatrix = "health_disk"
```
cgrinds, line 23: can be moved into const block above

grafana/dashboards/cmode/health.json
cgrinds, line 1: Few dashboard comments:

## Top row

Maybe something like this?
`
This dashboard requires ONTAP 9.6+ and the REST collector. Two actions are required to use this dashboard: <br>1. Enable the REST collector in your `harvest.yml` config<br>2. Enable the EMS collector in your `harvest.yml` config for EMS events<br>
`

## Disk table
We should make sure this table and the one used on the Disk dashboard are similar everywhere it makes sense. Same column names, same order? The Disk dashboard does not show the shelf, serial number, and model. Maybe they both should have those or neither should.

Hardikl (dismissed):
Looks good.

grafana/dashboards/cmode/health.json
```diff
@@ -0,0 +1,3219 @@
...
+        "sort": 1,
+        "tagValuesQuery": "",
+        "tagsQuery": "",
+        "type": "query",
+        "useTags": false
+      },
+      {
+        "current": {
+          "selected": false,
+          "text": "Provides EMS events selected Time Range configured in ems.yaml for EMS collector",
```
cgrinds, line 3177: maybe: The EMS collector collects EMS events defined in your `ems.yml` and this panel shows the ones with emergency severity

grafana/dashboards/cmode/health.json
```diff
@@ -0,0 +1,3219 @@
...
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 35
+      },
+      "id": 247,
+      "panels": [
+        {
+          "datasource": "${DS_PROMETHEUS}",
+          "description": "Health monitor events occurred in selected time range",
```
cgrinds, line 2973: Health monitor events that occurred in the selected time range

grafana/dashboards/cmode/health.json
```diff
@@ -3185,11 +3185,11 @@
...
           {
             "selected": true,
-            "text": "Provides EMS events selected Time Range configured in ems.yaml for EMS collector",
-            "value": "Provides EMS events selected Time Range configured in ems.yaml for EMS collector"
+            "text": "The EMS collector gathers EMS events as defined in your ems.yml file. This panel specifically displays events with emergency severity that occurred within a selected time range.",
+            "value": "The EMS collector gathers EMS events as defined in your ems.yml file. This panel specifically displays events with emergency severity that occurred within a selected time range."
           }
         ],
-        "query": "Provides EMS events selected Time Range configured in ems.yaml for EMS collector",
+        "query": "The EMS collector gathers EMS events as defined in your ems.yml file. This panel specifically displays events with emergency severity that occurred within a selected time range.",
```
cgrinds, line 3192: The EMS collector gathers EMS events as defined in your ems.yml file. This panel displays events with emergency severity that occurred within the selected time range.

cgrinds (commented):
looks good - two small change to the ems text:
- Drop specifically
- Change a => the

cgrinds (approved):

PR #1899, opened 2023-04-05
refactor: rename OntapS3 perf metrics to ontaps3_svm

cgrinds:


conf/zapiperf/default.yaml
```diff
@@ -48,7 +47,8 @@ objects:
   NFSv41:                 nfsv4_1.yaml
   NFSv42:                 nfsv4_2.yaml
   NFSv4:                  nfsv4.yaml
-#  SMB2:                   smb2.yaml
+  OntapS3_SVM:            ontap_s3_svm.yaml
```
Hardikl, line 50: underscode(_) would be needed in left side name ?

Hardikl: we do have in zapi as well `Status_7mode`, both way would be fine

Hardikl (approved):

Hardikl (approved):

PR #1903, opened 2023-04-05
fix: correct CI logs

rahulguptajss:


integration/test/dashboard_json_test.go
```diff
@@ -202,14 +202,14 @@ func (suite *DashboardJsonTestSuite) TestJsonExpression() {
 	}
 
 	if isRestFailed {
-		assert.Fail(suite.T(), "Rest Test validation is failed. Pls check logs above. Count of Missing Rest counters %d", len(perfErrorInfoList))
+		assert.Fail(suite.T(), fmt.Sprintf("Rest Test validation is failed. Pls check logs above. Count of Missing Rest counters %d", len(perfErrorInfoList)))
```
cgrinds, line 205: This var is called `perfErrorInfoList` but this log is about REST. Is this var used for more than REST or should it be renamed?

Hardikl (approved):

cgrinds (approved):

PR #1908, opened 2023-04-07
fix: log time drift between nodes in ems collector

Hardikl:
Testing is in-progress with the vsim.

### DIFF
--- a/cmd/collectors/ems/ems.go
+++ b/cmd/collectors/ems/ems.go
@@ -249,9 +249,10 @@ func (e *Ems) InitCache() error {
 
 func (e *Ems) getClusterTime() (time.Time, error) {
 	var (
-		err         error
-		records     []gjson.Result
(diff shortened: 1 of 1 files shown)


Hardikl:
Fix #1367

cmd/collectors/ems/ems.go
```diff
@@ -275,10 +276,14 @@ func (e *Ems) getClusterTime() (time.Time, error) {
 				continue
 			}
 			clusterTime = t
-			break
+			dateOfEachNode = append(dateOfEachNode, currentClusterDate.String())
```
cgrinds, line 279: Since we're already converting the string to time, it might be better to calculate the millisecond or seconds difference between each node and the first one. 

Accumulate that difference, and if it is greater than a predefined threshold, log the warning. That will allow some minor difference instead of requiring exact timestamps on each node.

PR #1925, opened 2023-04-11
fix: restrict exemplar flag in dashboards

rahulguptajss:


cmd/tools/grafana/dashboard_test.go
```diff
@@ -487,6 +487,24 @@ func TestIDIsBlank(t *testing.T) {
 		})
 }
 
+func TestExemplarTrue(t *testing.T) {
```
cgrinds, line 490: minor: but maybe `TestExemplarIsFalse` is better since it states the desired state instead of the problem state

cmd/tools/grafana/dashboard_test.go
```diff
@@ -487,6 +487,24 @@ func TestIDIsBlank(t *testing.T) {
...
+	visitDashboards(
+		[]string{"../../../grafana/dashboards/cmode", "../../../grafana/dashboards/storagegrid"},
+		func(path string, data []byte) {
+			checkExemplarTrue(t, path, data)
+		})
+}
+
+func checkExemplarTrue(t *testing.T, path string, data []byte) {
+
+	gjson.ForEachLine(string(data), func(line gjson.Result) bool {
```
cgrinds, line 500: probably better to use a different API since grafana dashboards are not json lines

cgrinds (approved):

PR #1941, opened 2023-04-14
feat: collect external_service_op perf object - part1

Hardikl:


Hardikl:
image

Hardikl:
Ready to review

cgrinds:
Nice job @Hardikl - since the object collector is disabled by default, this dashboard needs a row/note saying it needs to be enabled similar to Workload dashboards

integration/test/dashboard_json_test.go
```diff
@@ -40,7 +40,15 @@ var zapiCounterMap = map[string]struct{}{
 
 // restCounterMap are additional counters, above and beyond the ones from counterMap, which should be excluded from Rest
 var restCounterMap = map[string]struct{}{
-	"aggr_snapshot_inode_used_percent": {},
+	"aggr_snapshot_inode_used_percent":                {},
+	"external_service_op_num_not_found_responses":     {},
```
cgrinds, line 44: since these are included in `excludeCounters` they shouldn't be needed here too

integration/test/dashboard_json_test.go
```diff
@@ -49,6 +57,14 @@ var excludeCounters = []string{
 	"cluster_peer",
 	"efficiency_savings",
 	"ems_events",
+	"external_service_op_num_not_found_responses",
```
cgrinds, line 60: could simplify to `external_service_op_num_` and `external_service_op_request_` unless that excludes things that should be checked

Hardikl:
Couple of issues were identified as many more labels were included and started with plugin go way, then identified which labels would be sufficient for this. Then tried to solve same by moving them to template LabelAgent plugin, but these external service ops are across cluster, so just svm, service_name and operation won't be giving the exact result.

That's where I need to put changes in plugin go only with unique key as: `datacenter, cluster, svm, service_name, operation`.

Not a clean way of doing this, but this is what I found.

integration/test/dashboard_json_test.go
Hardikl: Done

integration/test/dashboard_json_test.go
Hardikl: Done

PR #1945, opened 2023-04-17
feat: Harvest should include a SG and ONTAP fabricpool dashboard

cgrinds:


Hardikl:
`ONTAP Cluster` and `SG Cluster` as variable would be giving little confusion here and while reading the query `cluster=~"$SGCluster"` it gets added.

image

how about this,
`ONTAP Cluster` --> `Cluster`, same as other dashboards
`SG Cluster` --> `Grid`, which is easy to interpret that's used only storagegrid metrics

grafana/dashboards/storagegrid/fabricpool.json
```diff
@@ -0,0 +1,1951 @@
...
+      "datasource": "${DS_PROMETHEUS}",
+      "description": "",
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "palette-classic"
+          },
+          "custom": {
+            "axisLabel": "",
+            "axisPlacement": "auto",
```
Hardikl, line 1391: image

These both the graphs are at policy level and then applying topk on policy. It took while to get this after few iteration of drop down. 

May be description of both panels would be helpful as topk=1 or 2 won't give 1 or 2 records but more which is based on the policy selected.

PR #1955, opened 2023-04-18
ci: simplify counter validation

cgrinds:


integration/test/dashboard_json_test.go
```diff
@@ -352,18 +325,30 @@ func FindStringBetweenTwoChar(stringValue string, startChar string, endChar stri
 }
 
 func hasDataInDB(query string) bool {
-	timeNow := time.Now().Unix()
-	queryUrl := fmt.Sprintf("%s/api/v1/query?query=%s&time=%d",
-		data.PrometheusURL, query, timeNow)
-	response, _ := utils.GetResponse(queryUrl)
-	value := gjson.Get(response, "data.result")
-	return value.Exists() && value.IsArray() && (len(value.Array()) > 0)
+	const waitFor = 6 * time.Minute
```
Hardikl, line 328: This 6 min is max of 2 Harvest poll time or some other calculation ?
If that's the case, do we need this at global const?

PR #1956, opened 2023-04-19
feat: collect lif counters

rahulguptajss:


Hardikl:
Minor: `At Home` values are true/false, do we want same as SM Yes/No ?

Hardikl (approved):
looks good.

Hardikl (approved):

PR #1979, opened 2023-04-21
feat: doctor should validate custom.yaml files

cgrinds:


cmd/tools/doctor/doctor_test.go
```diff
@@ -102,3 +102,52 @@ func loadTestConfig() {
...
 }
+
+func TestCustomYamlIsValid(t *testing.T) {
+	valid := checkCustomYaml("testdata/conf1")
+	if valid.isValid {
+		t.Error(`want isValid=false`)
+	}
+	if len(valid.invalid) != 1 {
+		t.Errorf(`want 1 invalid, got %d`, len(valid.invalid))
+	}
```
Hardikl, line 113: This would not be needed as it covered in below test cases.

Hardikl (approved):

PR #1984, opened 2023-04-24
fix: correct some mistakes in ems.yaml

cgrinds:
feat: sort ems.yaml

### DIFF
--- a/conf/ems/9.6.0/ems.yaml
+++ b/conf/ems/9.6.0/ems.yaml
@@ -11,14 +11,149 @@ exports:
   - index                                   => index
 
 events:
+  - name: LUN.destroy
+    exports:
+      - parameters.lun_path                 => lun_path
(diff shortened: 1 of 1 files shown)


conf/ems/9.6.0/ems.yaml
```diff
@@ -72,15 +207,15 @@ events:
...
       - parameters.volumeName               => volume
+      - parameters.volumeUuid               => volume_uuid
       - parameters.vserverName              => svm
       - parameters.vserverUuid              => svm_uuid
 
   - name: arw.vserver.state
     matches:
       - name: op
-        value: enabled
+        value: disabled
```
Hardikl, line 218: change here also else CI test might fail: https://github.com/NetApp/harvest/blob/main/integration/test/alert/alert.go#L27

Hardikl (approved):

PR #1993, opened 2023-04-25
ci: improve ems alert logging and event generation

cgrinds:


integration/test/alert/alert.go
```diff
@@ -93,8 +100,13 @@ func GetEmsAlerts(dir string, fileName string) ([]string, []string, []string) {
 func GenerateEvents(emsNames []string, nodeScopedEms []string) []string {
 	supportedEms := make([]string, 0)
 	var jsonValue []byte
-	addr, user, pass, nodeName := GetPollerDetail()
-	url := "https://" + addr + "/api/private/cli/event/generate"
+	err := conf.LoadHarvestConfig(installer.HarvestConfigFile)
+	poller, err2 := conf.PollerNamed(TestClusterName)
+	dc1, err3 := conf.PollerNamed("dc1")
```
Hardikl, line 105: only password we fetched from dc1, right ?

integration/test/bookend_ems_test.go
```diff
@@ -31,52 +32,54 @@ var nodeScopedResolvingEmsList = []string{
...
-
 	for _, issuingEms := range supportedIssuingEms {
-		// If the issuingEms did not exit before, then ignore the test-case.
+		// If the issuingEms did not exist before, then ignore the test-case.
 		if oldAlertsData[issuingEms] > 0 {
-			v := oldAlertsData[issuingEms] - newAlertsData[issuingEms]
-			if v < 1 {
-				foundBookendEms = append(foundBookendEms, issuingEms)
+			v := newAlertsData[issuingEms] - oldAlertsData[issuingEms]
+			if v > 1 {
```
Hardikl, line 78: This would not be as per our test case.
If we have issuing ems(ex. volume.offline) count as 3[oldAlertsData] earlier and after we fire resolving ems(volume.online), the issuing ems count would be reduced to 2[newAlertsData]
So, v = 2-3 = -1

Condition would be,
```
if v >= 0 {
error log
}
```

PR #2046, opened 2023-05-05
doc: update Docker Compose generation process to remove binary dependencies

rahulguptajss:


Hardikl (dismissed):

docs/install/containers.md
```diff
@@ -144,19 +149,21 @@ To upgrade Harvest:
    This is needed since the new version may contain new templates, dashboards, or other files not included in the Docker
    image.
 
-2. Copy your existing `harvest.yml` into the new Harvest directory created in step #1.
+2. [Stop all containers](#stop-all-containers). If you change directories, be sure to shut down any running Harvest containers from the previous directory.
```
cgrinds, line 152: We should change the `Stop all containers` section to include more words about this. Something like this:

> If you see this error [insert error] when trying to stop your containers, it is probably because you are running `docker-compose down` from a different directory than `up` was run. You can still stop and remove the old containers by running: `docker ps -qa | xargs docker stop | xargs docker rm` Note: This will stop and remove all running containers including any non-Harvest ones.

container/onePollerPerContainer/README.md
```diff
@@ -19,9 +19,13 @@ and [Dockerhub](https://hub.docker.com/r/rahulguptajss/harvest).
 If you want to create a separate container for each poller in your `harvest.yaml` file, download the latest version of Harvest and run 
 
 ```
-bin/harvest generate docker --output docker-compose.yml
+docker run --rm \
```
cgrinds, line 22: `--output docker-compose.yml` instead per chat

cgrinds (approved):

PR #2084, opened 2023-05-12
build: use go workspaces instead of replace

cgrinds:


cgrinds:
Closing until I figure out how to fix this - I suspect CodeQL does not understand Go workspaces yet.

```bash
go: -mod may only be set to readonly when in workspace mode, but it is set to "vendor"
	Remove the -mod flag to use the default readonly value,
	or set GOWORK=off to disable workspace mode.
```

PR #2120, opened 2023-06-01
feat: autoresolve ems would export metric value as 0 and autoresolve=true label

Hardikl:
image


cgrinds:
We should update the documentation to define the autoresolve label

Hardikl:
> We should update the documentation to define the autoresolve label

Added doc changes for this.

docs/configure-ems.md
```diff
@@ -111,7 +111,8 @@ The EMS event template parameters are explained below along with an example for
       named `LUN.online` is received.
     - `resolve_after` (optional, Go duration, default = 28 days) resolve the issuing EMS after the specified duration
       has elapsed (`672h` = `28d`).
-      If the bookend pair is not received within the `resolve_after` duration, the issuing EMS event expires.
+      If the bookend pair is not received within the `resolve_after` duration, the Issuing EMS event expires. Harvest
```
cgrinds, line 114: Maybe this

If the bookend pair is not received within the `resolve_after` duration, the issuing EMS event expires. When that happens, Harvest will mark the event as auto resolved by adding the `autoresolve=true` label to the issuing EMS event.

cmd/collectors/ems/ems.go
```diff
@@ -652,7 +643,16 @@ func (e *Ems) HandleResults(result []gjson.Result, prop map[string][]*emsProp) (
 			count += instanceLabelCount
 		}
 	}
-	return m, count
+
```
cgrinds, line 646: Can we calculate `instanceCount` in the above loops above, so we don't have to loop through all instances of all matrices again here?

cgrinds (commented):
will the logging issues we discussed in earlier meeting be handled in this PR or a different one?

cmd/collectors/ems/ems.go
Hardikl: It's added here mainly for `autoresolve` case.
In that case, we haven't get any records from ONTAP, so it won't run the above loop, but still we want to export 1 instance which would be resolved based on time.

Hardikl:
> will the logging issues we discussed in earlier meeting be handled in this PR or a different one?

Logging issue is resolved in this PR at 404-407 lines. 
It will be logged only exportable instance count not all belongs to e.Matrix.

Hardikl:
Matches label counts when multiple ems with same name exist would be handled in next PR

PR #2132, opened 2023-06-12
feat: support type label for volume for backward compatibility

rahulguptajss:


cmd/collectors/zapiperf/plugins/volume/volume.go
```diff
@@ -15,22 +15,41 @@ import (
 
 type Volume struct {
 	*plugin.AbstractPlugin
+	historicalLabels bool
```
cgrinds, line 18: can be removed, not used

cmd/collectors/restperf/plugins/volume/volume.go
```diff
@@ -11,18 +11,37 @@ import (
 
 type Volume struct {
 	*plugin.AbstractPlugin
+	historicalLabels bool
```
cgrinds, line 14: can be removed not used

cgrinds:
To modify your existing template make the following two changes
1. In the `plugins` section, replace the `Volume` line with
```
  - Volume:
      historicalLabels: true
```
2. In the `export_options instance_keys` section replace `style` with `type`

Example
```yaml
...
plugins:
  - Aggregator:
    - node
  - Volume:
      historicalLabels: true

export_options:
  instance_keys:
    - aggr
    - node
    - type
    - svm
    - volume
```

PR #2140, opened 2023-06-16
feat: volume clone info for ZAPI

rahulguptajss:


cmd/collectors/zapi/plugins/volume/volume.go
```diff
@@ -20,84 +20,171 @@ type Volume struct {
...
+					v.Logger.Error().Err(err).Str("metric", "clone_split_estimate").Msg("add metric")
+					continue
+				}
+			}
+
+			// splitEstimate is 4KB blocks, Convert to bytes as in REST
+
+			var splitEstimateBytes float64
+			if splitEstimateBytes, err = strconv.ParseFloat(vc.splitEstimate, 64); err != nil {
+				v.Logger.Error().Err(err).Str("metric", "clone_split_estimate").Str("value", vc.splitEstimate).Msg("parse metric")
```
cgrinds, line 134: minor: since the metric is static this could be rewritten as
`v.Logger.Error().Err(err).Str("clone_split_estimate", vc.splitEstimate).Msg("parse clone_split_estimate")`

cmd/collectors/zapi/plugins/volume/volume.go
```diff
@@ -20,84 +20,171 @@ type Volume struct {
...
+				continue
+			} else {
+				splitEstimateBytes = splitEstimateBytes * 4 * 1024
+			}
+			if err = splitEstimate.SetValueFloat64(volume, splitEstimateBytes); err != nil {
+				v.Logger.Error().Err(err).Str("metric", "clone_split_estimate").Str("value", vc.splitEstimate).Msg("set metric")
```
cgrinds, line 140: similar comment from above

cmd/collectors/zapi/plugins/volume/volume.go
```diff
@@ -20,84 +20,171 @@ type Volume struct {
...
-				my.Logger.Error().Err(err2).Msg("Failed to collect aggregate-disk mapping data")
+				v.Logger.Error().Err(err2).Msg("Failed to collect aggregate-disk mapping data")
 			}
 		}
 		// update aggrsMap based on disk data and addr disk mapping
-		my.updateAggrMap(disks, aggrDiskMap)
+		v.updateAggrMap(disks, aggrDiskMap)
+	}
+
+	volumeCloneMap, err := v.getVolumeCloneInfo()
```
Hardikl, line 94: Do we want to call this zapi at each poll or as other zapi call happens at plugin frequency to update map but still update volume labels at each poll from cache map ?

cgrinds (commented):
looks good, two minor logging comments

Hardikl (approved):

PR #2163, opened 2023-06-27
fix: correct the flashpool panel units

Hardikl:
Harvest 1.6 panels:
image

Harvest 2 panels:
image

It's for panels layout and unit comparison not for value comparison.


Hardikl:
Cluster `rtp-sa-cl05` has flashpool details.
image

cgrinds (approved):

PR #2182, opened 2023-07-05
refactor: eliminate usages of time.Sleep in test code

cgrinds:


cmd/collectors/commonutils.go
```diff
@@ -169,8 +169,9 @@ func SetNameservice(nsDB, nsSource, nisDomain string, instance *matrix.Instance)
 }
 
 // IsTimestampOlderThanDuration - timestamp units are micro seconds
-func IsTimestampOlderThanDuration(timestamp float64, duration time.Duration) bool {
-	return time.Since(time.UnixMicro(int64(timestamp))) > duration
+// The `begin` argument lets us virtualize time without requiring sleeps in test code
```
Hardikl, line 172: minor: may be name of var would be other than `begin` as it seems first time, but we actually are using last time - first time.

Hardikl (approved):

PR #2188, opened 2023-07-06
fix: ciphers query was giving wrong result in promQl

Hardikl:
Issue: Due to this below prom query, it would show 2 records in table, which is wrong in merge as it should be each row per cluster.

image

Fix: moved that regex based flag to plugin and added as label, and directly consume that label in query.

### DIFF
--- a/cmd/collectors/zapi/plugins/svm/svm.go
+++ b/cmd/collectors/zapi/plugins/svm/svm.go
@@ -13,6 +13,7 @@ import (
 	"github.com/netapp/harvest/v2/pkg/errs"
 	"github.com/netapp/harvest/v2/pkg/matrix"
 	"github.com/netapp/harvest/v2/pkg/tree/node"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -31,7 +32,7 @@ type SVM struct {
 	nisInfo        map[string]string
 	cifsEnabled    map[string]bool
 	nfsEnabled     map[string]string
-	sshData        map[string]string
+	sshData        map[string]ciphers
 	iscsiAuth      map[string]string
 	iscsiService   map[string]string
 	fpolicyData    map[string]fpolicy
@@ -53,6 +54,11 @@ type cifsSecurity struct {
 	cifsNtlmEnabled string
 	smbEncryption   string
 	smbSigning      string
+}
+
+type ciphers struct {
+	value     string
+	insecured string
 }
 
 func New(p *plugin.AbstractPlugin) plugin.Plugin {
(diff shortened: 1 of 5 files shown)


cmd/collectors/zapi/plugins/svm/svm.go
```diff
@@ -508,10 +515,17 @@ func (my *SVM) GetSSHData() (map[string]string, error) {
 		return nil, errs.New(errs.ErrNoInstance, "no records found")
 	}
 
+	compile, err := regexp.Compile("(.*)_cbc.*")
```
cgrinds, line 518: better to move outside the function into a var and use `MustCompile` so we only pay the cost to compile the regex once

cmd/collectors/zapi/plugins/svm/svm.go
```diff
@@ -508,10 +515,17 @@ func (my *SVM) GetSSHData() (map[string]string, error) {
...
+	if err != nil {
+		my.Logger.Error().Err(err).Msg("Failed to compile regex: (.*)_cbc.*")
+	}
+
 	for _, sshData := range result {
 		svmName := sshData.GetChildContentS("vserver-name")
 		sshList := sshData.GetChildS("ciphers").GetAllChildContentS()
-		sshMap[svmName] = strings.Join(sshList, ",")
+		ciphersVal := strings.Join(sshList, ",")
```
cgrinds, line 526: do we need a deterministic order for the joined list? Typically, we sort before joining

cmd/collectors/zapi/plugins/svm/svm.go
```diff
@@ -55,6 +56,11 @@ type cifsSecurity struct {
 	smbSigning      string
 }
 
+type ciphers struct {
```
cgrinds, line 59: maybe a better named struct would be this?
```go
type sshInfo struct {
	ciphers     string
	insecured string or isInsecure
}
```

Hardikl: added

cmd/collectors/zapi/plugins/svm/svm.go
Hardikl: we don't have that sorted order use-case for this, we just need list of supported ciphers and other boolean value(insecured) if it has `_cbc` weak ciphers.

cmd/collectors/zapi/plugins/svm/svm.go
Hardikl: Done

cmd/collectors/zapi/plugins/svm/svm.go
Hardikl: sorted the list before joining

cgrinds (approved):

PR #2204, opened 2023-07-12
feat: encapsulate fetchToken method in storagegrid

rahulguptajss:


cmd/collectors/storagegrid/rest/client.go
```diff
@@ -380,27 +385,28 @@ func (c *Client) fetchTokenWithAuthRetry() error {
...
+// fetchToken should not be used directly. Use fetchTokenWithAuthRetry instead.
+func (ci *clientInternal) fetchToken() error {
 	var (
 		err      error
 		req      *http.Request
 		response *http.Response
 		body     []byte
 	)
-	u, err := url.JoinPath(c.baseURL, c.APIPath, "authorize")
+	u, err := url.JoinPath(ci.client.baseURL, ci.client.APIPath, "authorize")
```
Hardikl, line 403: may be get client first as below and then use same for all places?
`c := ci.client`

Hardikl (approved):

PR #2211, opened 2023-07-18
feat: AFF250 Power calculation

rahulguptajss:


cmd/collectors/zapi/plugins/sensor/sensor.go
```diff
@@ -338,5 +344,17 @@ func (my *Sensor) calculateEnvironmentMetrics(data *matrix.Matrix) ([]*matrix.Ma
 			}
 		}
 	}
+
+	if len(whrSensors) > 0 {
+		var whrSensorsStr string
+		for _, v := range whrSensors {
+			if v != nil {
```
cgrinds, line 351: i don't think v can be nil so check not needed. Alternatively you could change `whrSensors` to `make(map[string]sensorValue)` too

Hardikl (approved):

grafana/dashboards/cmode/power.json
```diff
@@ -92,6 +98,36 @@
...
+      "datasource": "${DS_PROMETHEUS}",
+      "gridPos": {
+        "h": 2,
+        "w": 24,
+        "x": 0,
+        "y": 1
+      },
+      "id": 101,
+      "options": {
+        "content": "If an ONTAP system has embedded shelves, it may not have power and temperature information for shelf panels. This is because these details are calculated at the cluster level. Therefore, it is expected that you may not see power and temperature information for shelf panels in such systems.",
```
cgrinds, line 116: maybe clearer?

If an ONTAP system has **embedded** shelves, you may not see power and temperature information for those shelves since power and temperature are calculated at the cluster level.

cgrinds (approved):

PR #2238, opened 2023-07-26
feat: Harvest should fetch certificates via a script

cgrinds:


pkg/auth/auth.go
```diff
@@ -181,37 +255,151 @@ func (c *Credentials) GetPollerAuth() (PollerAuth, error) {
...
+	}
+
+	var pathPrefix string
+	certPath := poller.SslCert
+	keyPath := poller.SslKey
+
+	if certPath == "" {
+		o := &options.Options{}
+		options.SetPathsAndHostname(o)
+		pathPrefix = path.Join(o.HomePath, "cert/", o.Hostname)
```
Hardikl, line 323: these 3 lines could be outside this if block as it would be used by below if case also.

cmd/collectors/storagegrid/rest/client.go
```diff
@@ -100,49 +97,10 @@ func New(poller *conf.Poller, timeout time.Duration, c *auth.Credentials) (*Clie
...
-		transport = &http.Transport{
-			Proxy: http.ProxyFromEnvironment,
-			TLSClientConfig: &tls.Config{
-				Certificates:       []tls.Certificate{cert},
-				InsecureSkipVerify: useInsecureTLS}, //nolint:gosec
-		}
-	} else {
-		username := pollerAuth.Username
-		password := pollerAuth.Password
-		client.username = username
```
Hardikl, line 134: we need to retain this line for username same as rest/client.go 89 line

PR #2269, opened 2023-08-09
feat: added table description for cluster compliance

Hardikl:
image

### DIFF
--- a/conf/rest/9.10.0/volume.yaml
+++ b/conf/rest/9.10.0/volume.yaml
@@ -15,7 +15,7 @@ counters:
   - ^is_svm_root                                  => svm_root
   - ^snaplock.type                                => snaplock_type
   - ^snapshot_policy.name                         => snapshot_policy
-  - ^space.snapshot.autodelete.enabled            => snapshot_autodelete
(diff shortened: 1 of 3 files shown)


cgrinds:
Few comments:

- `Cluster has active certificate` => `Cluster has active certificate(s)`
- `Cluster has global fips enabled` => `Cluster has global FIPS enabled`
- `Cluster uses https for ASUP` => `Cluster uses HTTPS for autosupport`
- `Cluster has default admin user in locked state` => `Default admin user is locked`
- `Cluster has Remote shell disabled` => `Cluster's remote shell is disabled`
- `Cluster hasn't used md5 algorithm` => `Cluster does not use MD5 algorithm`
- `Cluster has secure ssh` => `Cluster has SSH enabled`
- `Cluster has configured 3 NTP servers` => `Cluster has configured three NTP servers`
- `Cluster has cluster peer encrypted` => `Cluster peers use encryption`?
- `Cluster has configured for automatic updates` => `Cluster has enabled automatic updates`

cgrinds (approved):

PR #2326, opened 2023-08-23
feat: ONTAP permission errors should be logged as errors

cgrinds:


cmd/poller/collector/collector.go
```diff
@@ -350,22 +350,24 @@ func (c *AbstractCollector) Start(wg *sync.WaitGroup) {
 						Msg("no metrics of object on system, entering standby mode")
 				// not an error we are expecting, so enter failed or standby state
 				default:
+					if errors.Is(err, errs.ErrPermissionDenied) {
+						c.Schedule.SetStandByMode(task, 1*time.Hour)
+						c.Logger.Error().Err(err).Str("task", task.Name).Msg("Entering standby mode")
```
Hardikl, line 355: better to add `Permission denied` in logs for more detail.

Hardikl (approved):

PR #2333, opened 2023-08-29
feat: ONTAP Power calculation for embedded Shelf

rahulguptajss:


cmd/collectors/commonutils.go
```diff
@@ -23,6 +23,37 @@ var validUnits = map[string]bool{
 	"W*hr":  true,
 }
 
+type embedShelf struct {
+	model, moduleType string
+}
+
+// Reference https://kb.netapp.com/onprem/ontap/hardware/FAQ%3A_How_do_shelf_product_IDs_and_modules_in_ONTAP_map_to_a_model_of_a_shelf_or_storage_system_with_embedded_storage
```
cgrinds, line 30: Maybe something like this?
```go
// Reference https://kb.netapp.com/onprem/ontap/hardware/FAQ%3A_How_do_shelf_product_IDs_and_modules_in_ONTAP_map_to_a_model_of_a_shelf_or_storage_system_with_embedded_storage
// There are two ways to identify embedded disk shelves:
// 1. The shelf's module type ends with E
// 2. The shelf is listed in the link above
var embeddedShelves = map[embedShelf]bool{
	embedShelf{"FS424-12", "IOM12F"}:   true,
}

func IsEmbedShelf(model string, moduleType string) bool {
	model = strings.ToUpper(model)
	moduleType = strings.ToUpper(moduleType)

    // if module type ends with E
	if strings.HasSuffix(moduleType, "E") {
		return true
	}

    return embeddedShelves[embedShelf{model, moduleType}]
}
```

cmd/collectors/zapi/plugins/shelf/shelf.go
```diff
@@ -55,81 +56,84 @@ func (my *Shelf) Init() error {
 		return err
 	}
 
-	my.query = "storage-shelf-environment-list-info"
+	if !my.client.IsClustered() {
```
cgrinds, line 60: did you mean to add the rest of the method to this conditional block?

if so, better to
if my.client.IsClustered() {
return nil
}
and out-dent the block

cgrinds (approved):

PR #2341, opened 2023-09-04
doc: add details about hidden_fields and filter for Rest Collector

rahulguptajss:


Hardikl (dismissed):

docs/configure-rest.md
```diff
@@ -113,6 +157,22 @@ The display name of a counter can be changed with `=>` (e.g., `space.block_stora
 
 Counters that are stored as labels will only be exported if they are included in the `export_options` section.
 
+The `counters` section allows you to specify `hidden_fields` and `filter` parameters. Please find the detailed explanation below.
+
+##### `hidden_fields`
+
+This is exclusively supported by the Rest Collector. The `hidden_fields` are specific fields that aren't inherently included in requests and must be explicitly requested by the user. When used, these fields are appended to the URL fields parameter.
```
cgrinds, line 164: Maybe this is clearer?

The `filter` is used to constrain the data returned by the endpoint, allowing for more targeted data retrieval. The filtering uses ONTAP's REST record filtering. The example above asks ONTAP to only return records where a volume's name matches `*harvest*`. 

If you're familiar with ONTAP's REST record filtering, the [example](#template-example) above would become `name=*harvest*` and appended to the final URL like so:

```
https://CLUSTER_IP/api/storage/volumes?fields=*,anti_ransomware.state,space&name=*harvest*
```

Refer to the ONTAP API specification, sections: `query parameters` and `record filtering`, for more details.

cgrinds (approved):

PR #2381, opened 2023-09-26
perf: Rest client allocs improvements

rahulguptajss:
This is very minor hence no profiling is done.

### DIFF
--- a/cmd/tools/rest/client.go
+++ b/cmd/tools/rest/client.go
@@ -208,9 +208,6 @@ func (c *Client) invokeWithAuthRetry() ([]byte, error) {
 
 		defer c.printRequestAndResponse(restReq, innerBody)
 
-		if innerErr != nil {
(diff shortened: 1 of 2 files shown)


cmd/tools/rest/client.go
```diff
@@ -206,16 +206,12 @@ func (c *Client) invokeWithAuthRetry() ([]byte, error) {
...
-		// read response body
-		if body, err = io.ReadAll(response.Body); err != nil {
-			return nil, err
-		}
-		defer c.printRequestAndResponse(restReq, body)
+		defer c.printRequestAndResponse(restReq, innerBody)
 
-		if err != nil {
-			return nil, err
+		if innerErr != nil {
```
cgrinds, line 211: not needed since the value of innerErr has not changed since line 182

Hardikl (approved):

PR #2396, opened 2023-10-02
perf: remove extra dict wrapper

cgrinds:
Reduces matrix size and removes unnecessary indirection


Hardikl (commented):
Other dict changes are fine but confuse about logging changes.

PR #2407, opened 2023-10-06
perf: avoid byte slice to string conversion

rahulguptajss:


Hardikl (dismissed):

pkg/api/ontapi/zapi/client.go
```diff
@@ -551,11 +551,11 @@ func (c *Client) TraceLogSet(collectorName string, config *node.Node) {
...
-	if response != nil {
-		res = string(response)
-	}
 	if req != "" {
+		res := "<nil>"
+		if response != nil {
+			res = string(response)
+		}
 		c.Logger.Info().
 			Str("Request", req).
```
cgrinds, line 560: did you try `Bytes("Response", response).` to avoid the conversion?

cgrinds (approved):

PR #2419, opened 2023-10-11
feat: add flexgroup to fabricpool panels

rahulguptajss:


Hardikl (approved):

cmd/collectors/fabricpool.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+func GetFlexGroupFabricPoolMetrics(dataMap map[string]*matrix.Matrix, object string, opName string, includeConstituents bool, l *logging.Logger) (*matrix.Matrix, error) {
+	var (
+		err error
+	)
+
+	data := dataMap[object]
+	opsKeyPrefix := "temp_"
+	re := regexp.MustCompile(`^(.*)__(\d{4})$`)
```
cgrinds, line 18: move outside function so we don't recompile every call

cmd/collectors/fabricpool.go
```diff
@@ -0,0 +1,140 @@
...
+			if cache.GetInstance(key) == nil {
+				fg, _ := cache.NewInstance(key)
+				fg.SetLabels(maps.Clone(i.GetLabels()))
+				fg.SetLabel("volume", match[1])
+			}
+			i.SetExportable(includeConstituents)
+		}
+	}
+
+	l.Logger.Debug().Msgf("extracted %d flexgroup volumes", len(cache.GetInstances()))
```
cgrinds, line 39: better to use Int and Msg instead of Msgf

cmd/collectors/fabricpool.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// create summary
+	for _, i := range data.GetInstances() {
+		if match := re.FindStringSubmatch(i.GetLabel("volume")); len(match) == 3 {
+			// instance key is svm.flexgroup-volume
+			key := i.GetLabel("svm") + "." + match[1]
+
+			fg := cache.GetInstance(key)
+			if fg == nil {
+				l.Logger.Error().Str("key", key).Msgf("instance not in local cache")
```
cgrinds, line 49: Msg instead of Msgf

cmd/collectors/fabricpool.go
```diff
@@ -0,0 +1,140 @@
...
+			}
+
+			for mkey, m := range data.GetMetrics() {
+				if !m.IsExportable() && m.GetType() != "float64" {
+					continue
+				}
+
+				fgm := cache.GetMetric(mkey)
+				if fgm == nil {
+					l.Logger.Error().Str("key", mkey).Msgf("metric not in local cache")
```
cgrinds, line 60: Msg not Msgf

cmd/collectors/fabricpool.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+
+	// normalize latency values
+	for _, i := range cache.GetInstances() {
+		if !i.IsExportable() {
+			continue
+		}
+		for mkey, m := range cache.GetMetrics() {
+			if m.IsExportable() && strings.HasPrefix(mkey, "cloud_bin_op_latency_average") {
```
cgrinds, line 120: This is code is being used by both zapi and rest plugins, which have different metric names, but we should be able to determine the correct metric name higher in this method and directly get the metric instead of looping over all metrics.
`cache.GetMetrics()[cloudBinOpLatencyAverage]`

cgrinds: right, buckets

cgrinds (approved):

PR #2420, opened 2023-10-11
feat: smb2 restperf counters

rahulguptajss:
The distinction between ZapiPerf and RestPerf is the additional `type` label in ZapiPerf, which is derived from parsing the `instance_uuid`. `instance_uuid` is missing in RestPerf. However, based on the dataset available in our clusters, it appears that the `type` label may not be necessary.

**ZapiPerf:**
`smb2_tree_connect_latency{datacenter="VSIM",cluster="sti62nscluster-1",node="sti62-vsim-ucs140d",svm="vs0",type="kernel"} 0`

**RestPerf:**
`smb2_session_setup_latency{datacenter="VSIM",cluster="sti62nscluster-1",node="sti62-vsim-ucs140d",svm="vs0"} 0`

### DIFF
--- a/conf/restperf/9.14.1/smb2.yaml
+++ b/conf/restperf/9.14.1/smb2.yaml
@@ -1,0 +1,45 @@
+name:                     SMB2
+query:                    api/cluster/counter/tables/smb2
+object:                   smb2
+
+counters:
+  - ^^id
+  - ^node.name                         => node
+  - ^svm.name                          => svm
+  - close_latency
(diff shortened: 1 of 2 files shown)


cgrinds (approved):

cgrinds:
Needs a release note since the RestPerf template does not have exact parity with the ZapiPerf template due to ONTAP impl

Hardikl (approved):

PR #2435, opened 2023-10-19
feat: adding restperf object_store_server template

Hardikl:


Hardikl:
Template changes are done.
This PR would be finished after the template testing will be done in latest cluster.

Hardikl:
image

image

PR #2442, opened 2023-10-24
doc: example auto-support payload should include REST example

cgrinds:


autosupport/README.md
```diff
@@ -24,2314 +24,4229 @@ An example payload sent by Harvest looks like this. You can see exactly what Har
...
-      "Name": "Zapi",
-      "Query": "volume-get-iter",
-      "BatchSize": "500",
-      "ClientTimeout": "180",
-      "Schedules": [
-        {
-          "Name": "instance",
-          "Schedule": "600s"
-        },
-        {
```
Hardikl, line 195: These zapi payloads are removed from here, as I couldn't find them in right side changes ?

PR #2445, opened 2023-10-25
feat: QoS fixed Utilization % panels

rahulguptajss:


Hardikl (approved):

Hardikl:
Minor: generally we don't show datacenter column in other tables. 
Do we want to show in both the tables here ?

PR #2484, opened 2023-11-13
feat: add changelog sample alerts

rahulguptajss:


container/prometheus/alert_rules.yml
```diff
@@ -74,4 +74,31 @@ groups:
...
+      description: "Snapmirror [{{ $labels.relationship_id }}] has [{{$value}}] lag time (in secs)"
+
+  # Object created. Refer https://netapp.github.io/harvest/latest/plugins/#changelog-plugin for more details.
+  - alert: Object Created
+    expr: change_log{op="create"} > 0
+    labels:
+      severity: "info"
+    annotations:
+      summary: "{{ $labels.object }} [{{ $labels.volume }}] created"
+      description: "{{ $labels.object }} [{{ $labels.volume }}] created"
```
Hardikl, line 86: not all object would have volume label.
May be target Alert for volume only with additional filter of `object=volume`

cgrinds (approved):

Hardikl (approved):

PR #2524, opened 2023-12-04
doc: add docker logging configuration guide

rahulguptajss:


docs/install/containers.md
```diff
@@ -108,6 +107,12 @@ Bring everything up :rocket:
 docker-compose -f prom-stack.yml -f harvest-compose.yml up -d --remove-orphans
 ```
 
+### Note on Docker Logging Configuration
+
+By default, Docker uses the `json-file` logging driver which does not limit the size of the logs. This can lead to issues if your system runs out of space. Docker provides several options for logging configuration, including different logging drivers and options for log rotation.
```
cgrinds, line 112: Maybe it is clearer to say 
"This can cause your system to run out of disk space."
instead of
"This can lead to issues if your system runs out of space."

docs/install/containers.md
```diff
@@ -108,6 +107,12 @@ Bring everything up :rocket:
...
+
+We recommend reviewing the [Docker logging documentation](https://docs.docker.com/config/containers/logging/configure/) to understand how to configure logging to suit your needs.
+
```
cgrinds, line 115: Docker recommends using the `local` driver to prevent disk-exhaustion. More details can be found in [Docker logging documentation](https://docs.docker.com/config/containers/logging/configure/).

cgrinds (approved):

PR #2535, opened 2023-12-06
fix: handling shelf_new_status for 7mode

Hardikl:
Instead of generating new Matrix for new instances in go plugin, Updating the `data` Matrix itself with extra instances generated from plugin, So that `Label_agent` plugin can be applied as is.

### DIFF
--- a/cmd/collectors/zapi/plugins/shelf/shelf.go
+++ b/cmd/collectors/zapi/plugins/shelf/shelf.go
@@ -18,7 +18,7 @@ const BatchSize = "500"
 type Shelf struct {
 	*plugin.AbstractPlugin
 	data                map[string]*matrix.Matrix
-	shelfData           *matrix.Matrix
+	shelfMetrics        *matrix.Matrix
 	instanceKeys        map[string]string
 	instanceLabels      map[string]map[string]string
 	shelfInstanceKeys   []string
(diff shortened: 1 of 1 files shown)


Hardikl:
Tested against the 7mode zapi response data shared by customer.

cmd/collectors/zapi/plugins/shelf/shelf.go
```diff
@@ -222,13 +223,18 @@ func (my *Shelf) handle7Mode(data *matrix.Matrix, result []*node.Node) ([]*matri
...
 
-	my.shelfData.SetGlobalLabels(data.GetGlobalLabels())
-	for _, instance := range data.GetInstances() {
-		instance.SetExportable(false)
+	// Purge instances generated from template and updated data metrics from shelfMetrics
+	data.PurgeInstances()
+	for metricName, m := range my.shelfMetrics.GetMetrics() {
+		_, err := data.NewMetricFloat64(metricName, m.GetName())
+		if err != nil {
+			my.Logger.Error().Err(err).Msg("add metric")
```
cgrinds, line 235: should we continue when there is an error?

Hardikl: Mostly we do error out where error occurred while instance creation, as nothing would be possible further that,
But for metrics creation or setting value in plugin, if error occurred then we logged as error and go ahead as it's impacting only that metric.

Hardikl:
**Before PR changes:**
shelf_labels:
image

shelf_new_status:
image

shelf_status_reads_attempted:
image

shelf_fan_labels:
image

shelf_fan_rpm:
image



**After the PR changes:**
shelf_labels:
image

shelf_new_status:
image

shelf_status_reads_attempted:
image

shelf_fan_labels:
image

shelf_fan_rpm:
image

Issue #2541, opened 2023-12-08
StorageGRID collectors should support only_cluster_instance


cgrinds:
### Discussed in https://github.com/NetApp/harvest/discussions/2540

Thanks to @santosh725 for raising.

Hardikl:
Without `only_cluster_instance: true` flag the SG metrics haven't been exported.

```
harvest % ./bin/poller -p sg-tudor -o Health --promPort 14001
2024-02-13T14:53:32+05:30 INF poller/poller.go:215 > Init Poller=sg-tudor asup=false confPath=conf config=harvest.yml configPath=harvest.yml daemon=false debug=false homePath= hostname=hardikl-mac-0 logLevel=info logPath=/var/log/harvest/ profiling=0 promPort=18001 version="harvest version 24.02.1313-v23.11.0 (commit febb62c5) (build date 2024-02-13T13:36:45+0530) darwin/amd64"
2024-02-13T14:53:32+05:30 INF poller/poller.go:239 > started in foreground Poller=sg-tudor pid=27686
2024-02-13T14:53:34+05:30 INF collector/helpers.go:84 > best-fit template Poller=sg-tudor collector=StorageGrid:Health path=conf/storagegrid/11.6.0/health.yaml v=11.7.0
2024-02-13T14:53:34+05:30 INF storagegrid/storagegrid.go:83 > initialized Poller=sg-tudor collector=StorageGrid:Health
2024-02-13T14:53:34+05:30 INF poller/poller.go:357 > Autosupport scheduled. Poller=sg-tudor asupSchedule=24h
2024-02-13T14:53:34+05:30 INF poller/poller.go:366 > poller start-up complete Poller=sg-tudor
2024-02-13T14:53:34+05:30 INF prometheus/httpd.go:40 > server listen Poller=sg-tudor exporter=prometheus1 url=http://:18001/metrics
2024-02-13T14:53:34+05:30 INF poller/poller.go:523 > updated status, up collectors: 1 (of 1), up exporters: 1 (of 1) Poller=sg-tudor
2024-02-13T14:53:35+05:30 INF collector/collector.go:585 > Collected Poller=sg-tudor apiMs=245 calcMs=0 collector=StorageGrid:Health exportMs=0 instances=0 instancesExported=0 metrics=0 metricsExported=0 parseMs=0 pluginMs=0 pollMs=245 zBegin=1707816214846
```

Tested in main with commit febb62c5

Template used for testing.

```
name: Health
query: grid/health
[...]

PR #2542, opened 2023-12-08
fix: StorageGRID collectors should support only_cluster_instance

cgrinds:
Fixes: #2541


Hardikl (dismissed):

Hardikl:
Does it worth it to add here https://netapp.github.io/harvest/23.11/configure-storagegrid/ that we do support `only_cluster_instance ` for storage grid?

PR #2548, opened 2023-12-13
fix: intermittent latency spike

rahulguptajss:
Because of this issue, harvest latency metrics may intermittently have spikes. These spikes are delta values, not average values.

### DIFF
--- a/pkg/matrix/matrix.go
+++ b/pkg/matrix/matrix.go
@@ -350,16 +350,18 @@ func (m *Matrix) Delta(metricKey string, prevMat *Matrix, logger *logging.Logger
 			prevIndex := prevInstance.index
 			if curMetric.record[currIndex] && prevRecord[prevIndex] {
 				curMetric.values[currIndex] -= prevRaw[prevIndex]
+				curCooked := curMetric.values[currIndex]
 				// Sometimes ONTAP sends spurious zeroes or values less than the previous poll.
+				// The condition also checks if the current cooked metric (curCooked) is not zero.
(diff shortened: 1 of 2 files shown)


Hardikl (approved):

pkg/matrix/matrix.go
```diff
@@ -350,16 +350,18 @@ func (m *Matrix) Delta(metricKey string, prevMat *Matrix, logger *logging.Logger
 			prevIndex := prevInstance.index
 			if curMetric.record[currIndex] && prevRecord[prevIndex] {
 				curMetric.values[currIndex] -= prevRaw[prevIndex]
+				curCooked := curMetric.values[currIndex]
 				// Sometimes ONTAP sends spurious zeroes or values less than the previous poll.
+				// It also ensures that the current cooked metric (curCooked) is not zero when either the current raw metric (curRaw) or the previous raw metric (prevRaw[prevIndex]) is zero.
+				// A non-zero curCooked under these conditions indicates an issue with the current or previous poll.
 				// Detect and don't publish negative deltas or the subsequent poll will show a large spike.
```
cgrinds, line 357: Maybe this is clearer?

// Sometimes ONTAP sends spurious zeroes or values less than the previous poll.
// Detect these cases and don't publish them, otherwise the subsequent poll will have large spikes.
// Ensure that the current cooked metric (curCooked) is not zero when either the current raw metric (curRaw) or the previous raw metric (prevRaw[prevIndex]) is zero.
// A non-zero curCooked under these conditions indicates an issue with the current or previous poll.

cgrinds (approved):

PR #2550, opened 2023-12-13
doc: Steps to create custom grafana dashboard

rahulguptajss:


Hardikl (dismissed):
looks good.

cgrinds:
We should add a few sentences near the top along the lines of:

Instead of creating a new dashboard, it may be faster to clone and modify an existing dashboard. Alternatively you can copy/paste an existing dashboard's panel into your new dashboard.

Hardikl (approved):

PR #2553, opened 2023-12-14
feat: use modifier for topk to improve svm dashboard performance

rahulguptajss:


Hardikl (approved):
Tested with 2.33 prometheus and data looks same.

PR #2555, opened 2023-12-14
fix: hide idle metric and max to auto for cpu_domain_busy

Hardikl:
Fixing in this panel:
1. Changes max from 100 to auto as this panel can go beyond 100%.
2. metric `idle` show as hidden(grayed) as default.


Hardikl:
> Looks good. As we discussed, Let's check cpu percentage used in other dashboards and see if we need to make any changes there.

Checked the other dashboards and usage of `cpu_domain_busy` and `cpu_busy` metrics, So far wherever we are using them, we are showing within 100% as expected.

PR #2568, opened 2024-01-02
refactor: Update Aggr dashboard to sync with SM

Hardikl:
**Previous:**
image

image

**With Changes:**
image

image

**System Manager For reference:**
image

### DIFF
--- a/grafana/dashboards/cmode/aggregate.json
+++ b/grafana/dashboards/cmode/aggregate.json
@@ -438,7 +438,7 @@
           "refId": "A"
         }
       ],
-      "title": "Used Space",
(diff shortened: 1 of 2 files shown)


cgrinds:
To better match SM, would it make more sense to rename Harvest's "Used Space" stat to "Used and Reserved" space?

Hardikl:
**Previous:**
image

image



**With Changes:**
image

image


**System Manager For reference:**
image

Hardikl:
> To better match SM, would it make more sense to rename Harvest's "Used Space" stat to "Used and Reserved" space?

Yes, Added the same for aggr and cluster dashboard.

PR #2569, opened 2024-01-02
doc: bump the minimum Prometheus version to 2.33

cgrinds:
Prometheus 2.33+ is required to use the PromQL modifier `@`.

See:
- https://discord.com/channels/855068651522490400/1184394042268717087
- https://prometheus.io/blog/2021/02/18/introducing-the-@-modifier/


Hardikl (approved):
We can change version here https://github.com/NetApp/harvest/blob/main/integration/test/installer/prometheus.go#L22 also for our CI

PR #2573, opened 2024-01-03
feat: RestPerf Disk Plugin should Support Metric Customization

Hardikl:
RestPerf Disk Plugin now supports Metric Customization via template changes.

This will support multi instance key as well
```
- paths => multipath:
    storage-shelf-path-info:
      - ^^controller => node
      - ^^initiator-wwpn => wwpn
```

### DIFF
--- a/conf/restperf/9.12.0/disk.yaml
+++ b/conf/restperf/9.12.0/disk.yaml
@@ -45,7 +45,43 @@ plugins:
   - Max:
       - node<>node_disk_max
       - aggr<>aggr_disk_max ...
-  - Disk
+
+  - Disk:
+      objects:
+        - fans => fan:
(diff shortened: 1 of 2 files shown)


conf/restperf/9.12.0/disk.yaml
```diff
@@ -45,7 +45,43 @@ plugins:
   - Max:
       - node<>node_disk_max
       - aggr<>aggr_disk_max ...
-  - Disk
+
+  - Disk:
+      objects:
+        - fans => fan:
+            - ^^id => fan_id
```
cgrinds, line 52: probably better to align on `=>` like we do in the counters section

Hardikl: Done

PR #2575, opened 2024-01-04
feat: add filter support for ZapiPerf collector

rahulguptajss:


docs/configure-zapi.md
```diff
@@ -189,3 +189,76 @@ exporters, the following parameters can be defined:
 * `instances_keys` (list): display names of labels to export with each data-point
 * `instance_labels` (list): display names of labels to export as a separate data-point
 * `include_all_labels` (bool): export all labels with each data-point (overrides previous two parameters)
+
+### Filter
+
+This guide provides detailed instructions on how to use the `filter` feature in ZapiPerf.
```
cgrinds, line 195: We should say something about why you would want to use filter. Maybe something like this:

This guide provides instructions on how to use the `filter` feature in ZapiPerf. Filtering is useful when you want to query a subset of instances. Say you have a small number of high-value volumes that you want Harvest to collect performance metrics from every five seconds. You can't collect all volumes, every five seconds, because that would be too expensive. Filtering lets you create a separate template that includes only the high-value volumes.

docs/configure-zapi.md
```diff
@@ -189,3 +189,76 @@ exporters, the following parameters can be defined:
...
+
+#### Objects (Excluding Workload)
+
+In ZapiPerf templates, you can set up filters under `counters`. Wildcards like * are useful if you don't want to specify all instances.
```
cgrinds, line 199: Please note, ONTAP Zapi filtering does not support regular expressions, only wildcard matching with * and ?.

cgrinds: Check if `?` works and if not remove that clause

docs/configure-zapi.md
```diff
@@ -189,3 +189,76 @@ exporters, the following parameters can be defined:
...
+
+For instance, to filter Volume performance instances by instance name where the name is `NS_svm_nvme` or contains `Test`, use the following configuration in ZapiPerf `volume.yaml` under `counters`:
```
cgrinds, line 201: volume instance of Volume

docs/configure-zapi.md
```diff
@@ -189,3 +189,76 @@ exporters, the following parameters can be defined:
...
+    - A pipe ("|") signifies an OR operation.
+    - The precedence order is AND first, followed by OR.
+
+    For instance, the filter string `instance_name=volA,vserver_name=vs1|vserver_name=vs2` translates to `(instance_name=volA && vserver_name=vs1) || (vserver_name=vs2)`.
+
+    This filter will return instances on Vserver `vs1` named `volA`, and all instances on Vserver `vs2`.
+
+#### Workload Templates
+
+Performance Workload Templates are treated differently in Harvest, hence they follow a unique approach. You can use filter fields provided via `qos-workload-get-iter` as shown below.
```
cgrinds, line 230: Performance workload templates require a different syntax because instances are retrieved from the `qos-workload-get-iter` ZAPI instead of `perf-object-instance-list-info-iter`. 

The `qos-workload-get-iter` ZAPI supports filtering on the following fields:

- workload-uuid
- workload-name
- wid
- category
- policy-group
- vserver
- volume
- lun
- file
- qtree
- read-ahead
- max-throughput
- min-throughput
- is-adaptive
- is-constituent

cgrinds (approved):

Hardikl (approved):

PR #2583, opened 2024-01-11
feat: flexcache monitoring

rahulguptajss:


cmd/collectors/zapiperf/zapiperf.go
```diff
@@ -57,6 +58,7 @@ const (
 	instanceKey   = "uuid"
 	batchSize     = 500
 	latencyIoReqd = 10
+	keyToken      = "?#"
```
Hardikl, line 61: why this is `?#`

Hardikl (approved):

PR #2586, opened 2024-01-15
ci: template validation for rest, restperf

Hardikl:


integration/test/counter_test.go
```diff
@@ -0,0 +1,184 @@
...
+		log.Fatal().Err(err).Str("poller", pollerName).Msgf("")
+	}
+	if poller.Addr == "" {
+		log.Fatal().Str("poller", pollerName).Msg("Address is empty")
+	}
+	timeout, _ := time.ParseDuration(rest2.DefaultTimeout)
+
+	if client, err = rest2.New(poller, timeout, auth.NewCredentials(poller, logging.Get())); err != nil {
+		log.Fatal().Err(err).Str("poller", pollerName).Msg("error creating new client")
+		os.Exit(1)
```
cgrinds, line 55: os.Exit is called by log.Fatal()

integration/test/counter_test.go
```diff
@@ -0,0 +1,184 @@
...
+		r := eachTemp(path, client.Cluster().GetVersion(), client)
+		for k, v := range r {
+			result[k] = v
+		}
+		return nil
+	})
+
+	if err != nil {
+		log.Fatal().Err(err).Msg("failed to read template:")
+		return nil
```
cgrinds, line 124: code never reached so can be removed

integration/test/counter_test.go
```diff
@@ -0,0 +1,184 @@
...
+	for _, countersDetail := range counters {
+		for _, counterDetail := range countersDetail {
+			href := rest2.NewHrefBuilder().
+				APIPath(counterDetail.api).
+				Fields(counterDetail.restCounters).
+				CounterSchema(counterDetail.perfCounters).
+				Build()
+
+			if _, err := collectors.InvokeRestCall(client, href, logging.Get()); err != nil {
+				log.Info().Msgf("%s", href)
```
cgrinds, line 79: Might be better to augment the error with the href like so and then remove the log.Info
`return fmt.Errorf("failed to invoke rest href=%s call: %w", href, err)`

integration/test/counter_test.go
```diff
@@ -0,0 +1,184 @@
...
+func invokeRestCall(client *rest2.Client, counters map[string][]counterData) error {
+	for _, countersDetail := range counters {
+		for _, counterDetail := range countersDetail {
+			href := rest2.NewHrefBuilder().
+				APIPath(counterDetail.api).
+				Fields(counterDetail.restCounters).
+				CounterSchema(counterDetail.perfCounters).
+				Build()
+
+			if _, err := collectors.InvokeRestCall(client, href, logging.Get()); err != nil {
```
cgrinds, line 78: do we need to check the response to see if any data is returned, or is it sufficient to only check that no error is returned?

Hardikl: Out of 3 cases, 
1. Error in call, 
2. Empty records(no objects exist in cluster), 
3. Proper response.

`InvokeRestCall` function can return err in first case, and that's we need to check as case 2 is valid with 0 records.

cgrinds: Thanks. My question was do we want these tests to pass if no records are returned from the cluster? I believe we do since these tests are meant to check that the template `counters` are valid ONTAP REST fields.

Might help to add a comment to the `TestCounters` function that spells out the purpose, maybe something like this:

> TestCounters extracts non-hidden counters from all of the rest and restperf templates and then invokes an HTTP GET for each api path + counters. Valid responses are status code = 200. Objects do not need to exist on the cluster, only the api path and counter names are checked.

conf/rest/9.12.0/netport.yaml
```diff
@@ -11,7 +11,6 @@ counters:
   - ^lag.active_ports.name                   => lag_port
   - ^lag.distribution_policy                 => lag_distribution_policy
   - ^lag.mode                                => lag_mode
-  - ^operational_speed                       => op_speed
```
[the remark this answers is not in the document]
Hardikl: image

PR #2606, opened 2024-01-25
doc: add FSA template description

rahulguptajss:


Hardikl (approved):

conf/rest/9.12.0/volume_analytics.yaml
```diff
@@ -6,14 +6,21 @@ counters:
...
+      - order_by=space.used desc  # Sort volumes in descending order based on used space
+      - max_records=20  # Limit the records to the top 20 volumes
 
 plugins:
+  # The 'VolumeAnalytics' plugin collects analytics data from directories within each volume.
   - VolumeAnalytics:
+      # 'MaxDirectoryCount' specifies the number of directories to collect per volume.
+      # In this case, data is collected from up to 100 directories, prioritizing those with the highest used bytes.
       - MaxDirectoryCount: 100
+      # The following sample URL gets called per volume in the VolumeAnalytics plugin:
```
cgrinds, line 22: Do you think something like the following would be clearer without tying the documentation to the exact implementation?

Using the plugin settings shown above, the VolumeAnalytics plugin will make an additional request for each volume. Those per-volume requests will look something like:
api/storage/volumes/{uuid}/files?return_records=true&type=directory&max_records=100

cgrinds (approved):

PR #2637, opened 2024-02-06
feat: improving Debug Log Clarity and Reducing Noise

rahulguptajss:


cmd/collectors/zapi/plugins/qtree/qtree.go
```diff
@@ -421,11 +438,16 @@ func (q *Qtree) handlingQuotaMetrics(quotas []*node.Node, cluster string, quotaI
 						q.Logger.Debug().Msgf("(%s) failed to parse value (%s): %v", attribute, value, err)
 					} else {
 						*numMetrics++
-						q.Logger.Debug().Msgf("(%s) added value (%s)", attribute, value)
+						q.Logger.Trace().
+							Str("attribute", attribute).
+							Interface("value", value).
```
cgrinds, line 443: lets not use Interface since it uses reflection, instead we can use Str("attrValue", attrValue)

cmd/collectors/zapiperf/plugins/disk/disk.go
```diff
@@ -874,9 +894,16 @@ func (d *Disk) handleCMode(shelves []*node.Node) ([]*matrix.Matrix, error) {
...
+									d.Logger.Debug().
+										Str("metricKey", metricKey).
+										Str("value", value).
+										Err(err).
+										Msg("failed to parse value")
 								} else {
-									d.Logger.Debug().Msgf("(%s) added value (%s)", metricKey, value)
+									d.Logger.Trace().
+										Str("metricKey", metricKey).
+										Interface("value", value).
```
cgrinds, line 905: same comment as before about Interface

cgrinds (commented):
nice! Thanks for changing the printf style ones

cgrinds (approved):

Hardikl (approved):

PR #2674, opened 2024-02-15
fix: honor absolute paths from the HARVEST_CONF environment variable

cgrinds:


Hardikl (dismissed):
changes are fine, just to add "" case.

PR #2715, opened 2024-02-26
feat: dashboards would work with grafana 10.x.x

cgrinds:


Hardikl:
Changes look good and handled most of the customer where they have legacy name, Added one test case for that.

Hardikl (approved):

PR #2747, opened 2024-03-14
feat: Harvest should handle ONTAP counter manager rejection errors

cgrinds:
The collector will retry after 30 to 60 seconds

### DIFF
--- a/cmd/collectors/restperf/restperf.go
+++ b/cmd/collectors/restperf/restperf.go
@@ -1544,9 +1544,6 @@ func (r *RestPerf) handleError(err error, href string) (map[string]*matrix.Matri
 		// the table or API does not exist. return ErrAPIRequestRejected so the task goes to stand-by
 		return nil, fmt.Errorf("polling href=[%s] err: %w", href, errs.New(errs.ErrAPIRequestRejected, err.Error()))
 	}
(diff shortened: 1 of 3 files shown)


cmd/collectors/restperf/restperf.go
```diff
@@ -1545,6 +1544,9 @@ func (r *RestPerf) handleError(err error, href string) (map[string]*matrix.Matri
 		// the table or API does not exist. return ErrAPIRequestRejected so the task goes to stand-by
 		return nil, fmt.Errorf("polling href=[%s] err: %w", href, errs.New(errs.ErrAPIRequestRejected, err.Error()))
 	}
+	// if errs.IsRestErr(err, errs.CMReject) {
+	// 	return nil, fmt.Errorf("CM reject href=[%s] err: %w", href, errs.New(errs.ErrNoInstance, err.Error()))
+	// }
```
Hardikl, line 1549: Is this required ?

Hardikl (approved):

cmd/collectors/restperf/restperf.go
cgrinds: good catch. This was from an earlier dev on this problem. Removed since the collector handles errs.CMReject

PR #2848, opened 2024-04-23
feat: adaptive qos policy used monitoring (Zapi)

rahulguptajss:
image

image

### DIFF
--- a/grafana/dashboards/cmode/workload.json
+++ b/grafana/dashboards/cmode/workload.json
@@ -65,7 +65,7 @@
   "gnetId": null,
   "graphTooltip": 1,
   "id": null,
-  "iteration": 1713794854200,
+  "iteration": 1713941923885,
   "links": [
     {
       "asDropdown": true,
@@ -1718,7 +1718,7 @@
         },
         {
           "datasource": "${DS_PROMETHEUS}",
-          "description": "The Adaptive QoS Utilized IOPs (%) panel displays the current usage of IOPs as a percentage of the adaptive threshold set by an adaptive Quality of Service (QoS) policy. This metric is calculated by dividing the current IOPs usage by the adaptive threshold and multiplying by 100. The panel is active only when the block size is set to values, excluding the `any`.",
+          "description": "The Adaptive QoS Utilized IOPs (%) panel displays the current usage of IOPs as a percentage of the adaptive threshold set by an adaptive Quality of Service (QoS) policy. This metric is calculated by dividing the current IOPs usage by the adaptive threshold and multiplying by 100. The panel shows data only if the block size is not set to `any`.",
           "fieldConfig": {
             "defaults": {
               "color": {
@@ -1804,7 +1804,7 @@
               "refId": "A"
             }
           ],
-          "title": "Top $TopResources QoS Bandwidth Utilization (%)",
+          "title": "Top $TopResources Adaptive QoS Bandwidth Utilization (%)",
           "transformations": [],
           "type": "timeseries"
         },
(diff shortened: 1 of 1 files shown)


pkg/util/util.go
```diff
@@ -460,3 +462,18 @@ func HandleArrayFormat(name string) string {
...
+		return 0, nil
+	}
+	bytes, err := strconv.ParseInt(s, 10, 64)
+	if err != nil {
+		return 0, err
+	}
+	return float64(bytes) / bytesInTB, nil
+}
+
+func ConvertStringToFloat64(s string) (float64, error) {
```
cgrinds, line 477: We should use strconv.ParseFloat(s, 64) instead of ConvertStringToFloat64(s). This can be removed

cmd/collectors/zapiperf/plugins/workload/workload.go
```diff
@@ -0,0 +1,438 @@
...
+	return nil
+}
+
+func (w *Workload) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix, *util.Metadata, error) {
+
+	var (
+		err error
+	)
+
+	// The ZapiPerf test triggers the workload plugin, so the client init has been done to the Run method instead of the plugin init.
```
cgrinds, line 75: Is this still needed? This code is not hit when I run `zapiperf_test.go`

cmd/collectors/zapiperf/plugins/workload/workload.go
```diff
@@ -0,0 +1,438 @@
...
+		request       *node.Node
+		err           error
+		qosadativeMap map[string]*collectors.QosAdaptive
+	)
+
+	qosadativeMap = make(map[string]*collectors.QosAdaptive)
+	request = node.NewXMLS("qos-adaptive-policy-group-get-iter")
+	request.NewChildS("max-records", batchSize)
+
+	// Fetching only admin SVMs
```
cgrinds, line 391: Doesn't this fetch all SVMs without a filter?

cmd/collectors/zapiperf/plugins/workload/workload.go
```diff
@@ -0,0 +1,438 @@
...
+			size := v.GetChildS("volume-space-attributes").GetChildContentS("size")
+			sizeUsed := v.GetChildS("volume-space-attributes").GetChildContentS("size-used")
+			volume := v.GetChildS("volume-id-attributes").GetChildContentS("name")
+			svm := v.GetChildS("volume-id-attributes").GetChildContentS("owning-vserver-name")
+			key := volume + token + svm
+			instance := params.data.GetInstance(params.volumeWorkloadMap[key])
+			if instance != nil {
+				err = w.processWorkloadInstance(instance, size, sizeUsed, params.qosPolicyMap, params.data)
+				if err != nil {
+					w.Logger.Error().Err(err).Str(key, key).Msg("Failed process workload")
```
cgrinds, line 228: Should key be quoted like so?
`w.Logger.Error().Err(err).Str("key", key).Msg("Failed process workload")`

cmd/collectors/zapiperf/plugins/workload/workload.go
```diff
@@ -0,0 +1,438 @@
...
+		err = data.GetMetric("used_ops_percent").SetValueFloat64(instance, usedPercent)
+		if err != nil {
+			return err
+		}
+	}
+
+	if policy.BlockSize != "any" {
+		totalDataBps, record := data.GetMetric("total_data").GetValueFloat64(instance)
+		if record {
+			// possible values are "4k", "8k", "16k", "32k", "64k", "128k"
```
cgrinds, line 362: Can you add a comment about whether these are base2 or base10?

cmd/collectors/adaptive.go
```diff
@@ -0,0 +1,64 @@
...
+	if policy.PeakIOPSAllocation == "used_space" {
+		peakAllowedIOPS = max(absoluteMinIOPS, sizeUsedTB*peakIOPS)
+	} else {
+		peakAllowedIOPS = max(absoluteMinIOPS, sizeTB*peakIOPS)
+	}
+
+	return math.Round(peakAllowedIOPS), nil
+}
+
+func CalculateThroughputPercent(totalDataBps, blockSize, peakIOPS float64) float64 {
```
cgrinds, line 51: Can you add a comment about base2 or base10 for these values?

PR #2850, opened 2024-04-23
chore: address code scanning issues in thirdparty

cgrinds:


cgrinds:
This is not needed, closing

Issue #2497, opened 2023-11-21
Adaptive policy dashboard


rahulguptajss:
This is a follow up story from #2267 #1426. We should create a adaptive policy panels similiar to QOS fixed % in workload dashboard.

faguayot:
Hello @rahulguptajss 

First at all thanks for all the information shared in a order and visual way.

I have some doubts related the  shared information. In this example that you expose,  we have asummed that the adaptative QoS policy is implemented on a `volume level` but for example in our particular case, we are applying the adaptative QoS in the `LUN level.` 
Additionally, we don't use the **throughput** for limitated the ceiling or floor (probably other clients do this). And the last thing I have seen is we don't have a block size for every specific QoS defined too, in our case `the block size is ANY `and the `Peak and Expected IOPS Allocation` are: **allocated-space**.

But all the information that I have shared is only our use case, so I don't know if all the calculation that you share could be apply and adapt depending on the different way of defining the QoS adaptative policy. 

I hope I have explained myself successfully.
Thanks.

faguayot:
Hello Rahul, 

I have download the nightly from the link that you share but I don't see the same informacion about % of Adaptative QoS.

![image](https://github.com/NetApp/harvest/assets/85484449/babecf89-7c0b-499d-b2e9-40b72e9a100d)

In the `QoS bandwidth` is normal since we don't configure and use but where it should appear the list of workload, cluster,etc I only see `%qos used, IOPS and Max IOPS`.

Thanks. It looks great in your image dashboard.

faguayot:
I have tried importing the JSON as a new dashboard but the result is the same,

![image](https://github.com/NetApp/harvest/assets/85484449/542fdbc1-3227-4a0f-a9df-7832c47a9e11)

faguayot:
Hello rahul, 

With this importation is working correctly.

![image](https://github.com/NetApp/harvest/assets/85484449/ef361a53-d9cc-4cb6-aa51-4104188eaab0)


> @faguayot Could you try re-importing the dashboard from [here](https://raw.githubusercontent.com/NetApp/harvest/3f8f1e7dd29ae8e56f5ad9f94a535792eabdee07/grafana/dashboards/cmode/workload.json) to see if it helps?

Hardikl:
Verified in 24.05 release branch at commit 74dcf38a872d2c38e633664314e0acc5b9f8d76b

Tested in grafana 8.3.4,
[...]

PR #2873, opened 2024-05-03
fix: adaptive qos table grafana 9 workaround

rahulguptajss:
Thanks to @faguayot for reporting!

### DIFF
--- a/grafana/dashboards/cmode/workload.json
+++ b/grafana/dashboards/cmode/workload.json
@@ -65,7 +65,7 @@
   "gnetId": null,
   "graphTooltip": 1,
   "id": null,
-  "iteration": 1714734444899,
+  "iteration": 1714973870447,
   "links": [
     {
       "asDropdown": true,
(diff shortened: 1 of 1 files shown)


cgrinds:
See https://github.com/NetApp/harvest/issues/2497#issuecomment-2092529318

As @faguayot discovered, using Grafana 9 or 10, the table does not show some of the columns. As discussed on Teams, this problem is likely a bug in Grafana. We don't see the problem with Grafana 8, 9, or 10 when we import this dashboard. This bug happens when using a Grafana outer join transform with table queries that use `label_join`s.

The workaround is:
- Use the Grafana transform `Rename by regex` to rename results that end with "space 1" by trimming the "space 1". The label `cluster 1` becomes `cluster`.
- Use the Grafana transform `Filter by name` to only include results that do not end with a space followed by one or more digits. E.g. include cluster, exclude cluster 2, cluster 3, etc.

Hardikl (approved):

PR #2874, opened 2024-05-03
fix: Handling index for quota

Hardikl:
Index will be look like this:
`index="umeng-aff300-01-02_trident_qtree_pool_trident_TIXRBILLKA.trident_pvc_19913841_a29f_4a54_8bc0_a3c1c4155826.", `

There are 2 functions in zapi and rest, one for handling `handlingHistoricalMetrics`, Which I intentionally haven't changed.


Hardikl:
Testing is in-progress.

Hardikl:
> Can we get rid of index totally?

As soon as I remove index in Quota Report dashboard, it's broken.
image


Evaluating other ways...

PR #2931, opened 2024-05-28
feat: linking dashboard volume, aggregate, svm part-1

Hardikl:


Hardikl:
`Volume` dashboard changes are done. 
Testing import/export workflow and other uid related cases.

Hardikl:
Changes covered so far:
1. `UID` would be non-empty and human readable form for all dashboards.
2. Folder name would be generic and don't have release/date in it, same behaviour as NABox.
3. No need to have `overwrite` flag as we always overwrite with same uid (will remove this arg later)
4. Volume dashboard all tables have links
5. Grafan Import, then export to local folder and then importing this local folder works fine.
6. Docker workflow testing is in-progress.

Hardikl:
Tested in `.127` system for docker workflow and it's working fine with new changes.

cmd/tools/grafana/dashboard_test.go
```diff
@@ -781,19 +781,11 @@ func checkExemplarIsFalse(t *testing.T, path string, data []byte) {
 	}
 }
 
-var uidExceptions = map[string]bool{
-	"cmode-details/volumeBySVM.json":    true,
-	"cmode-details/volumeDeepDive.json": true,
-}
-
-func checkUIDIsBlank(t *testing.T, path string, data []byte) {
+func checkUIDNonEmpty(t *testing.T, path string, data []byte) {
```
cgrinds, line 784: minor, but checkUIDNotEmpty might be better

cmd/tools/grafana/grafana.go
```diff
@@ -400,10 +398,10 @@ func initImportVars() {
 
 	// default behaviour
 	if opts.dir == "grafana/dashboards" && opts.serverfolder.name == "" {
-		m[filepath.Join(opts.dir, "cmode")] = &Folder{name: "Harvest-" + harvestRelease + "-cDOT"}
-		m[filepath.Join(opts.dir, "cmode", "details")] = &Folder{name: "Harvest-" + harvestRelease + "-cDOT Details"}
-		m[filepath.Join(opts.dir, "7mode")] = &Folder{name: "Harvest-" + harvestRelease + "-7mode"}
-		m[filepath.Join(opts.dir, "storagegrid")] = &Folder{name: "Harvest-" + harvestRelease + "-StorageGrid"}
+		m[filepath.Join(opts.dir, "cmode")] = &Folder{name: "Harvest-main-cDOT"}
```
cgrinds, line 401: If we're ditching versioned folders maybe `Harvest-cDOT`, `Harvest-StorageGrid`, etc. would be better?

cmd/tools/grafana/grafana.go
```diff
@@ -485,36 +483,17 @@ func importFiles(dir string, folder *Folder) {
...
-					data, err = sjson.SetBytes(data, "uid", []byte(""))
-					if err != nil {
-						fmt.Printf("error while updating the uid %s into dashboard %s, err: %+v", dashboardID, file.Name(), err)
-						continue
-					}
-				}
-			}
-		}
-
 		// If the dashboard has an id defined, change the id to empty string, unless overwrite was passed,
```
cgrinds, line 486: we can remove this code

cmd/tools/grafana/grafana.go
```diff
@@ -1165,7 +1144,7 @@ func addImportExportFlags(commands ...*cobra.Command) {
 		cmd.PersistentFlags().StringVarP(&opts.addr, "addr", "a", "http://127.0.0.1:3000", "Address of Grafana server (IP, FQDN or hostname)")
 		cmd.PersistentFlags().StringVarP(&opts.token, "token", "t", "", "API token issued by Grafana server for authentication")
 		cmd.PersistentFlags().BoolVarP(&opts.useHTTPS, "https", "S", false, "Use HTTPS")
-		cmd.PersistentFlags().BoolVarP(&opts.overwrite, "overwrite", "o", false, "Overwrite existing dashboard with same title")
+		cmd.PersistentFlags().BoolVarP(&opts.overwrite, "overwrite", "o", true, "Overwrite existing dashboard with same title")
```
cgrinds, line 1147: We should leave this argument with a default value of false. When the arg is set, we should log a warning saying something like: 'the overwrite flag is no longer used and will be removed in a future release. Please remove this flag from your command line invocation. When importing, dashboards are always overwritten.'

cgrinds (commented):
Since Grafana has a flat namespace and folders don't really exist perhaps we should simplify the uids like so:
```
harvest-cdot-aggregate                       => cdot-aggregate
harvest-cdot-details-volume-by-svm => cdot-volume-by-svm
harvest-storagegrid-overview             => storagegrid-overview
```

cmd/tools/grafana/dashboard_test.go
Hardikl: done

cmd/tools/grafana/grafana.go
Hardikl: yeah, I was just matching the behaviour with what our docker workflow currently doing.

cmd/tools/grafana/grafana.go
Hardikl: done

cmd/tools/grafana/grafana.go
Hardikl: Added the check.

Hardikl:
> Since Grafana has a flat namespace and folders don't really exist, perhaps we should simplify the uids like so:
> 
> ```
> harvest-cdot-aggregate             => cdot-aggregate
> harvest-cdot-details-volume-by-svm => cdot-volume-by-svm
> harvest-storagegrid-overview       => storagegrid-overview
> ```

Added the naming convention this way.

cgrinds (dismissed):

cgrinds (approved):

PR #2940, opened 2024-05-29
doc: update docker instructions

rahulguptajss:


docs/install/containers.md
```diff
@@ -129,8 +129,8 @@ at `http://IP_OF_PROMETHEUS:9090/targets`.
 
 ### Customize Prometheus's Retention Time
 
-By default, `prom-stack.yml` is configured for a 15-day data retention period.
-To increase this, for example, to one year, you can create a specific configuration file and make your changes there.
+By default, `prom-stack.yml` is configured for a 1-year data retention period.
```
cgrinds, line 132: one year

docs/install/containers.md
```diff
@@ -227,7 +227,8 @@ To upgrade Harvest:
 2. [Stop all containers](#stop-all-containers)
 
 3. Regenerate your `harvest-compose.yml` file by
-   running [harvest generate](#generate-a-docker-compose-for-your-pollers)
+   running [harvest generate](#generate-a-docker-compose-for-your-pollers).
+   This step is essential as it updates local copies of templates and dashboards, which are then mounted to the containers
```
cgrinds, line 231: Make sure you don't skip this step. It is essential as it updates local copies of templates and dashboards, which are then mounted to the containers. If this step is skipped, Harvest will run with older templates and dashboards which will likely cause problems.

PR #2943, opened 2024-05-29
fix: Rest templates should not have hyphon

rahulguptajss:


cmd/tools/template/template_test.go
```diff
@@ -345,6 +345,22 @@ func TestOverrideMetricsExist(t *testing.T) {
 	}, allTemplatesButEms...)
 }
 
+func TestNoHyphenInMetrics(t *testing.T) {
+	visitTemplates(t, func(path string, model Model) {
+		isRest := strings.Contains(path, "rest") || strings.Contains(path, "restperf")
```
cgrinds, line 350: maybe this is simpler

		if !strings.Contains(path, "rest") {
			return
		}

cgrinds (approved):

PR #2946, opened 2024-05-30
ci: fix container stop in ci

rahulguptajss:


cgrinds (approved):

integration/test/dashboard_json_test.go
```diff
@@ -28,17 +28,21 @@ var fileSet []string
...
+	"ontaps3_logical_used_size":                 {},
+	"ontaps3_size":                              {},
+	"ontaps3_object_count":                      {},
+	"ontaps3_used_percent":                      {},
+	"ontaps3_services_labels":                   {},
+	"ontaps3_policy_labels":                     {},
+	"volume_arw_status":                         {},
+	"volume_num_compress_fail":                  {},
+	"volume_num_compress_attempts":              {},
+	"volume_performance_tier_footprint":         {},
```
cgrinds, line 42: maybe add a comment that this is temporary while we sort out u2 and sar? Alternatively, include those details in the PR so we have a paper trail of why this change was made and an indication on when we can undo it

cgrinds (dismissed):

cgrinds (approved):

PR #2960, opened 2024-06-03
chore: move gopsutil to v4

rahulguptajss:


cgrinds:
Handled with cherry pick in #2961

PR #2976, opened 2024-06-10
feat: linking dashboards part-3

Hardikl:


Hardikl:
All the dashboards have been tested, but these (as their metrics would be created based on the condition): 
- Health
- Changelog Monitor
- MetroCluster

PR #2980, opened 2024-06-11
fix: fixing non-exported flexgroup instances error

Hardikl:


cgrinds:
Thanks to Ankur Madaan for reporting.

This is a regression introduced in #2936 that causes flexgroup volume performance metrics to be missing.

PR #3000, opened 2024-06-18
feat: Implementing certificate expiry detail in security dashboard

Hardikl:
image

### DIFF
--- a/conf/rest/9.12.0/certificate.yaml
+++ b/conf/rest/9.12.0/certificate.yaml
@@ -1,0 +1,24 @@
+
+name:             Certificate
+query:            api/security/certificates
+object:           certificate
+
+counters:
+  - ^^uuid
+  - ^name
+  - ^public_certificate      => certificatePEM
+  - ^scope                   => scope
+  - ^serial_number           => serial_number
+  - ^svm.name                => svm
+  - ^type                    => type
+  - expiry_time(timestamp)   => expiry_time
+
+export_options:
+  instance_keys:
+    - name
+    - svm
+    - uuid
+  instance_labels:
+    - scope
+    - serial_number
+    - type
--- a/grafana/dashboards/cmode/security.json
+++ b/grafana/dashboards/cmode/security.json
@@ -1901,13 +1901,165 @@
       "type": "stat"
     },
     {
+      "datasource": "${DS_PROMETHEUS}",
+      "description": "This panel displays Certificate expiration time.",
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "fixedColor": "transparent",
+            "mode": "fixed"
+          },
+          "custom": {
+            "align": "left",
+            "cellOptions": {
+              "type": "auto"
+            },
+            "filterable": true,
+            "inspect": false
+          },
+          "mappings": [],
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
+                "value": null
+              }
+            ]
+          },
+          "unitScale": true
+        },
+        "overrides": [
+          {
+            "matcher": {
+              "id": "byName",
+              "options": "Expire"
+            },
+            "properties": [
+              {
+                "id": "unit",
+                "value": "dateTimeFromNow"
+              }
+            ]
+          }
+        ]
+      },
+      "gridPos": {
+        "h": 7,
+        "w": 24,
+        "x": 0,
+        "y": 26
+      },
+      "id": 228,
+      "options": {
+        "cellHeight": "sm",
+        "footer": {
+          "countRows": false,
+          "fields": "",
+          "reducer": [
+            "sum"
+          ],
+          "show": false
+        },
+        "showHeader": true,
+        "sortBy": [
+          {
+            "desc": true,
+            "displayName": "Value #B"
+          }
+        ]
+      },
+      "pluginVersion": "10.3.1",
+      "targets": [
+        {
+          "datasource": "${DS_PROMETHEUS}",
+          "editorMode": "code",
+          "exemplar": false,
+          "expr": "certificate_labels{datacenter=~\"$Datacenter\", cluster=~\"$Cluster\", svm=~\"$SVM\"}",
+          "format": "table",
+          "hide": false,
+          "instant": true,
+          "interval": "",
+          "legendFormat": "",
+          "refId": "A"
+        },
+        {
+          "datasource": "${DS_PROMETHEUS}",
+          "editorMode": "code",
+          "exemplar": false,
+          "expr": "certificate_expiry_time{datacenter=~\"$Datacenter\", cluster=~\"$Cluster\", svm=~\"$SVM\"} * 1000",
+          "format": "table",
+          "hide": false,
+          "instant": true,
+          "interval": "",
+          "legendFormat": "",
+          "refId": "B"
+        }
+      ],
+      "title": "Certificates Detail",
+      "transformations": [
+        {
+          "id": "joinByField",
+          "options": {
+            "byField": "uuid",
+            "mode": "outer"
+          }
+        },
+        {
+          "id": "filterFieldsByName",
+          "options": {
+            "include": {
+              "names": [
+                "cluster",
+                "name",
+                "scope",
+                "svm",
+                "type",
+                "Value #B"
+              ]
+            }
+          }
+        },
+        {
+          "id": "organize",
+          "options": {
+            "excludeByName": {
+              "cluster 2": true,
+              "datacenter 2": true,
+              "isEncrypted": true,
+              "isHardwareEncrypted": true,
+              "name 2": true,
+              "svm 2": true
+            },
+            "includeByName": {},
+            "indexByName": {
+              "cluster": 0,
+              "name": 2,
+              "scope": 3,
+              "svm": 1,
+              "type": 4
+            },
+            "renameByName": {
+              "Value #B": "Expire",
+              "cluster": "Cluster",
+              "name": "Certificate Name",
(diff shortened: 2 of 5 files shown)


Hardikl:
image

image

cgrinds (commented):
Other comments shared in chat

grafana/dashboards/cmode/security.json
```diff
@@ -1900,14 +1900,185 @@
       ],
       "type": "stat"
     },
+    {
```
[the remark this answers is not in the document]
Hardikl: So, While checking the `security/certificates` Rest call, there are no certificates records for root svms. We are good to go here.

Also, I would be adding the `scope` field, which shows `cluster` or `svm` to help customer to see the scope of the certificate.

conf/rest/9.12.0/certificate.yaml
```diff
@@ -0,0 +1,18 @@
+
+name:             Certificate
+query:            api/security/certificates
+object:           certificate
+
+counters:
```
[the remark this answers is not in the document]
Hardikl: added.

conf/rest/9.12.0/certificate.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - ^expiry_time             => expiry_time
+  - ^name
+  - ^svm.name                => svm
+
+export_options:
+  instance_keys:
+    - name
+    - svm
+    - uuid
+  instance_labels:
```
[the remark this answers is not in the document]
Hardikl: As this is label, it's quite difficult to write alert, Let me explore to handle this.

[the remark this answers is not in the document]
Hardikl: I need to use metric instead of label for this: 
This is the alert query for certificates expiring within 1 month:
`0 < (certificate_expiry_time{} - time()) < (30*24*3600)`

I will add sample warning alert as well for reference.

grafana/dashboards/cmode/security.json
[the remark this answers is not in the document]
cgrinds: Are you saying:
a) that root SVMs do not have certificates or 
b) that ONTAP does not return certificates for root SVMs

I think you're saying that root SVM have certificates, but ONTAP is not returning them? If that's the case, we should check the expiry for root SVM certificates some other way

Hardikl:
image

Hardikl:
security_certificate.yml creates 2 metrics, one `security_certificate_labels_labels` with extra 2 labels and one metric `security_certificate_expiry_time` which never used in dashboard. 


```
harvest % curl -s http://localhost:13001/metrics | grep -Ev "#|metadata_" | grep security 
security_certificate_labels{certificateExpiryStatus="expired",certificateIssuerType="self_signed",cluster="umeng-aff300-01-02",datacenter="rest",name="umeng-aff300-01-02_172BB302639CAC9B",scope="cluster",serial_number="172BB302639CAC9B",svm="umeng-aff300-01-02",type="server",uuid="806b6fe6-6ef8-11ed-a5cd-00a098d39e12"} 1.0
security_certificate_expiry_time{cluster="umeng-aff300-01-02",datacenter="rest",uuid="806b6fe6-6ef8-11ed-a5cd-00a098d39e12"} 1701160909
harvest % curl -s http://localhost:13001/metrics | grep -Ev "#|metadata_" | grep security  | wc -l
       2
```

So, to re-use this existing template, only one change is required in dashboard where passing `certificateExpiryStatus!=\"\",certificateIssuerType!=\"\"` in `security_certificate_labels` where earlier it was used all extra filters would be removed from the template.

grafana/dashboards/cmode/security.json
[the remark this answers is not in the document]
Hardikl: So, there is little correction on my above note.
- Out of `admin` and `node` svm (which we treat as `root svm`), `admin` svm can have certificates
- But, It's cluster scope and not svm scope, (`scope` is only in Rest) which means we don't get svm name in certificate in Rest calls.
- We would do little more work to get those detail in Rest, which we are already doing in existing plugin, and we have this comment as well for that reference. 
`// Admin SVM certificate is cluster scoped, but the REST API does not return the SVM name in its response. Add here for ZAPI parity`
- As there are cluster scoped and svm scoped certificates in table, I need to remove the svm filter from query, 
So, even the SVM drop down have limited svms but this table shows certificates from all of them.
Screenshot from .127 system
image


Just to note, Above the table the stats count is showing those admin svm's certificates only(admin svm is unique in cluster) and not all of them.

cmd/collectors/zapi/plugins/certificate/certificate.go
```diff
@@ -98,18 +100,30 @@ func (my *Certificate) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix,
...
-				if serialNumber == adminVserverSerial {
-					certificateInstance.SetExportable(true)
+				if expiryTimeMetric = data.GetMetric("certificate-info.expiration-date"); expiryTimeMetric == nil {
+					my.Logger.Error().Msg("missing expiry time metric")
+					continue
+				}
+				if expiryTime, ok := expiryTimeMetric.GetValueFloat64(certificateInstance); ok {
+					// convert expiryTime from float64 to int64 and then to unix Time
+					unixTime = time.Unix(int64(expiryTime), 0)
+					certificateInstance.SetLabel("expiry_time", unixTime.UTC().String())
```
cgrinds, line 118: Use .Format(time.RFC3339) instead of String()

cmd/collectors/zapi/plugins/certificate/certificate.go
```diff
@@ -98,18 +100,30 @@ func (my *Certificate) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix,
 
 		// update certificate instance based on admin vaserver serial
 		for certificateInstanceKey, certificateInstance := range data.GetInstances() {
+			unixTime = time.Now()
```
cgrinds, line 103: Is this needed?

cmd/collectors/rest/plugins/certificate/certificate.go
```diff
@@ -87,20 +89,31 @@ func (my *Certificate) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix,
 
 		// update certificate instance based on admin vaserver serial
 		for _, certificateInstance := range data.GetInstances() {
+			unixTime = time.Now()
```
cgrinds, line 92: is this needed?

cmd/collectors/rest/plugins/certificate/certificate.go
```diff
@@ -87,20 +89,31 @@ func (my *Certificate) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix,
...
-					certificateInstance.SetExportable(true)
+				if expiryTimeMetric = data.GetMetric("expiry_time"); expiryTimeMetric == nil {
+					my.Logger.Error().Stack().Msg("missing expiry time metric")
+					continue
+				}
+
+				if expiryTime, ok := expiryTimeMetric.GetValueFloat64(certificateInstance); ok {
+					// convert expiryTime from float64 to int64 and then to unix Time
+					unixTime = time.Unix(int64(expiryTime), 0)
+					certificateInstance.SetLabel("expiry_time", unixTime.UTC().String())
```
cgrinds, line 106: Use .Format(time.RFC3339) instead of String()

cmd/collectors/zapi/plugins/certificate/certificate.go
Hardikl: Done

cmd/collectors/zapi/plugins/certificate/certificate.go
Hardikl: changed to else

cmd/collectors/zapi/plugins/certificate/certificate.go
```diff
@@ -96,20 +98,35 @@ func (my *Certificate) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix,
...
-					my.setCertificateIssuerType(certificateInstance, certificateInstanceKey)
-					my.setCertificateValidity(data, certificateInstance)
-				}
+			if !certificateInstance.IsExportable() {
+				continue
+			}
+			name := certificateInstance.GetLabel("name")
+			serialNumber := certificateInstance.GetLabel("serial_number")
+			svm := certificateInstance.GetLabel("svm")
+			CertType := certificateInstance.GetLabel("type")
```
cgrinds, line 109: certType instead of CertType

cmd/collectors/rest/plugins/certificate/certificate.go
```diff
@@ -85,22 +87,33 @@ func (my *Certificate) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix,
...
-					// Admin SVM certificate is cluster scoped, but the REST API does not return the SVM name in its response. Add here for ZAPI parity
-					certificateInstance.SetLabel("svm", adminVserver)
-					my.setCertificateIssuerType(certificateInstance)
-					my.setCertificateValidity(data, certificateInstance)
-				}
+			if !certificateInstance.IsExportable() {
+				continue
+			}
+			serialNumber := certificateInstance.GetLabel("serial_number")
+			CertType := certificateInstance.GetLabel("type")
```
cgrinds, line 96: certType instead of CertType

cmd/collectors/zapi/plugins/certificate/certificate.go
Hardikl: Done

PR #3018, opened 2024-06-26
fix: handle partial aggregation for flexgroup perf metrics

rahulguptajss:


cmd/collectors/volume.go
```diff
@@ -0,0 +1,208 @@
...
+							if err != nil {
+								logger.Error().Err(err).Msg("error")
+							}
+						}
+						err = fgm.SetValueFloat64(fg, fgv+prod)
+						if err != nil {
+							logger.Error().Err(err).Msg("error")
+						}
+					} else {
+						if recordFGFalse[key] == nil {
```
cgrinds, line 159: this is slightly more idiomatic and does one lookup instead of two (after creation)
```go
s, ok := recordFGFalse[key]
if !ok {
	s = set.New()
	recordFGFalse[key] = s
}
s.Add(fgm.GetName())
```

cgrinds (approved):
Looks good! One minor comment about a slightly more idiomatic map

Hardikl (approved):

Hardikl (approved):

PR #3019, opened 2024-06-28
feat: Harvest should remove Service Center metrics

rahulguptajss:


cmd/collectors/zapiperf/zapiperf.go
```diff
@@ -119,10 +119,27 @@ func (z *ZapiPerf) Init(a *collector.AbstractCollector) error {
...
 	}
 
+	if err := z.InitQOS(); err != nil {
+		return err
+	}
 	z.Logger.Debug().Msg("initialized")
 	return nil
 }
 
+func (z *ZapiPerf) InitQOS() error {
```
cgrinds, line 129: the method can be `func (z *ZapiPerf) InitQOS() {` since no error is returned

cgrinds (commented):
A small number of changes. Nice!

Hardikl (approved):

PR #3041, opened 2024-07-09
feat: enable ChangeLog plugin to monitor metric value change

rahulguptajss:


docs/plugins.md
```diff
@@ -748,6 +747,23 @@ Example of metric shape for object deletion:
 change_log{aggr="umeng_aff300_aggr2", cluster="umeng-aff300-01-02", datacenter="u2", index="2", instance="localhost:12993", job="prometheus", node="umeng-aff300-01", object="volume", op="delete", style="flexvol", svm="harvest", volume="harvest_demo"} 1698735708
 ```
 
+### Metric Value Changes
+
+When a metric value changes, the ChangeLog plugin will publish a metric with the following labels. Note that old and new values are not tracked to avoid impacting cardinality in Prometheus:
+
+| Label        | Description                                                                 |
+|--------------|-----------------------------------------------------------------------------|
+| object       | name of the ONTAP object that was changed                                   |
```
Hardikl, line 756: How to enabled the changelog for metric ?
We may need to add example where to add `read_latency` if want to monitor that metric.

does adding `read_latency` here would work?
```
plugins:
  - ChangeLog:
      track:
        - aggr
        - node
        - state
```

cmd/poller/plugin/changelog/changelog.go
```diff
@@ -243,12 +267,53 @@ func (c *ChangeLog) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix, *u
 	return matricesArray, nil, nil
 }
 
+// CompareMetrics compares the metrics of the current and previous instances
+func (c *ChangeLog) CompareMetrics(curMat *matrix.Matrix) map[string]map[string][2]float64 {
```
cgrinds, line 271: Since the cur and prev values are never used this can be simplified

cmd/poller/plugin/changelog/changelog.go
```diff
@@ -243,12 +267,53 @@ func (c *ChangeLog) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix, *u
...
+}
+
 // copyPreviousData creates a copy of the previous data for comparison
 func (c *ChangeLog) copyPreviousData(cur *matrix.Matrix) {
 	labels := c.changeLogConfig.PublishLabels
-	labels = append(labels, c.changeLogConfig.Track...)
+	var met []string
+	for _, t := range c.changeLogConfig.Track {
+		mKey := cur.DisplayMetricKey(t)
+		if mKey != "" {
```
cgrinds, line 305: minor but it's better to prefer the positive equality instead of the negative so
```go
if mKey == "" {
    labels = append(labels, t)
		} else {
    met = append(met, mKey)		
		}
```

cmd/poller/plugin/changelog/changelog.go
```diff
@@ -243,12 +267,53 @@ func (c *ChangeLog) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix, *u
...
+			met = append(met, mKey)
+		} else {
+			labels = append(labels, t)
+		}
+	}
 	labels = append(labels, "uuid")
-	c.previousData = cur.Clone(matrix.With{Data: true, Metrics: false, Instances: true, ExportInstances: false, Labels: labels})
+	withMetrics := false
```
cgrinds, line 312: withMetrics := len(met) > 0

cgrinds (approved):

PR #3066, opened 2024-07-25
feat: Support Oauth2.0 via credential script - phase1

Hardikl:


cmd/tools/rest/client.go
```diff
@@ -128,14 +129,20 @@ func (c *Client) GetRest(request string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	c.request.Header.Set("Accept", "application/json")
```
cgrinds, line 131: Should be on right side

cmd/tools/rest/client.go
```diff
@@ -128,14 +129,20 @@ func (c *Client) GetRest(request string) ([]byte, error) {
...
 	if err != nil {
 		return nil, err
 	}
-	if pollerAuth.Username != "" {
-		c.request.SetBasicAuth(pollerAuth.Username, pollerAuth.Password)
+	if pollerAuth.AuthToken != "" {
+		c.request.Header.Set("Authorization", "Bearer "+pollerAuth.AuthToken)
+		c.Logger.Debug().Msg("Using authToken from credential script")
+	} else {
+		c.request.Header.Set("Accept", "application/json")
```
cgrinds, line 140: not needed after line 131 is restored

cmd/tools/rest/client.go
Hardikl: reverted.

PR #3093, opened 2024-08-08
fix: remove threshold from quota rest template

Hardikl:


cmd/collectors/rest/plugins/quota/quota.go
```diff
@@ -25,6 +29,14 @@ func (q *Quota) Init() error {
 func (q *Quota) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix, *util.Metadata, error) {
 	data := dataMap[q.Object]
 
+	// Adding threshold metric once as it's not available from quota template
```
cgrinds, line 32: // The threshold metric does not exist in REST quota template, we are adding it to maintain parity with exported ZAPI metrics

cgrinds (dismissed):

cgrinds (approved):

PR #3111, opened 2024-08-16
perf: reduce allocs when reading credential files

cgrinds:
refactor: remove cyclic dependency in logging


cgrinds:
Before/after allocations of `ReadCredentialFile` after running for 1 hour.
| Result |   Mem  |
|--------|:------:|
| Before | 456 MB |
| After  | 3.5 MB |

130x improvement

Hardikl (approved):

PR #3124, opened 2024-08-27
fix: handled nameservices in svm zapi plugin

Hardikl:


cmd/collectors/rest/plugins/svm/svm.go
```diff
@@ -174,29 +174,29 @@ func (my *SVM) GetNSSwitchInfo(data *matrix.Matrix) (map[string]Nsswitch, error)
 
 	var (
 		vserverNsswitchMap map[string]Nsswitch
-		ns                 Nsswitch
 		ok                 bool
 	)
 
 	vserverNsswitchMap = make(map[string]Nsswitch)
+	replaceStr := strings.NewReplacer("[", "", "]", "", "\"", "", "\n", "", " ", "")
```
cgrinds, line 181: This can be built once at the package level instead of each time the method is called

Hardikl:
Validated with customer zapi response with this PR changes, Changes would be worked as expected.

PR #3132, opened 2024-09-02
feat: Top clients metrics collection

rahulguptajss:


cmd/collectors/restperf/plugins/volumetopclients/volumetopclients.go
```diff
@@ -0,0 +1,409 @@
...
+)
+
+var opMetrics = []string{
+	"ops",
+}
+var dataMetrics = []string{
+	"data",
+}
+
+type Volume struct {
```
cgrinds, line 42: Would a better name be VolumeTopClients or TopClients? Volume is a bit confusing since there are n volumes per cluster. This struct is really the plugin, which there will be one per poller instance. Volume makes it sound like there will be n per poller instance. On the other hand, I guess we do the same in volume.go for the Rest/Zapi volume plugin.

cmd/collectors/restperf/plugins/volumetopclients/volumetopclients.go
```diff
@@ -0,0 +1,409 @@
...
+	keyToken                 = "?#"
+	topClientReadOPSMatrix   = "volume_top_clients_read_ops"
+	topClientWriteOPSMatrix  = "volume_top_clients_write_ops"
+	topClientReadDataMatrix  = "volume_top_clients_read_data"
+	topClientWriteDataMatrix = "volume_top_clients_write_data"
+	defaultTopN              = 5
+	maxTopN                  = 50
+)
+
+var opMetrics = []string{
```
cgrinds, line 35: These slices only hold a single value which means they don't need to be slices. Do you anticipate that we will add other metrics to these? If not, maybe remove these and inline

cmd/collectors/restperf/plugins/volumetopclients/volumetopclients.go
```diff
@@ -0,0 +1,409 @@
...
+
+	if err := v.client.Init(5); err != nil {
+		return err
+	}
+
+	if maxVol := v.Params.GetChildContentS("MaxVolumeCount"); maxVol != "" {
+		if maxVolCount, err := strconv.Atoi(maxVol); err != nil {
+			v.maxVolumeCount = defaultTopN
+		} else {
+			v.maxVolumeCount = int(math.Min(float64(maxVolCount), float64(maxTopN)))
```
cgrinds, line 125: v.maxVolumeCount = min(maxVolCount, maxTopN)

cmd/collectors/restperf/plugins/volumetopclients/volumetopclients.go
```diff
@@ -0,0 +1,409 @@
...
+}
+
+func (v *Volume) collectMetricValues(data *matrix.Matrix, filteredDataInstances *set.Set) ([]MetricValue, []MetricValue, []MetricValue, []MetricValue) {
+	readOpsMetric := data.GetMetric("total_read_ops")
+	writeOpsMetric := data.GetMetric("total_write_ops")
+	readDataMetric := data.GetMetric("bytes_read")
+	writeDataMetric := data.GetMetric("bytes_written")
+
+	var readOpsList, writeOpsList, readDataList, writeDataList []MetricValue
+	for _, key := range filteredDataInstances.Slice() {
```
cgrinds, line 242: better to use Iter() instead of Slice() since Slice() will cause extra allocs that aren't necessary

cmd/collectors/restperf/plugins/volumetopclients/volumetopclients.go
```diff
@@ -0,0 +1,409 @@
...
+		var readOps, writeOps, readData, writeData float64
+		var ro, wo, rb, wb bool
+		if readOpsMetric != nil {
+			readOps, ro = readOpsMetric.GetValueFloat64(data.GetInstance(key))
```
cgrinds, line 246: data.GetInstance(key) is called four times in this loop. Would be better to call it once at the top of the loop and use the single instance

cmd/collectors/restperf/plugins/volumetopclients/volumetopclients.go
```diff
@@ -0,0 +1,409 @@
...
+		}
+		if wb && writeData > 0 {
+			writeDataList = append(writeDataList, MetricValue{key: key, value: writeData})
+		}
+	}
+	return readOpsList, writeOpsList, readDataList, writeDataList
+}
+
+func (v *Volume) getTopN(metricList []MetricValue) []MetricValue {
+	sort.Slice(metricList, func(i, j int) bool {
```
cgrinds, line 276: the docs suggest this
```go
slices.SortFunc(metricList, func(a, b MetricValue) int {
    return cmp.Compare(a.value, b.value)
})
```

cmd/collectors/restperf/plugins/volumetopclients/volumetopclients.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+	return volumes, svms
+}
+
+func (v *Volume) processTopClientsByMetric(volumes, svms *set.Set, matrixName, metric string, setMetricFunc func(*matrix.Matrix, *matrix.Instance, float64)) error {
+	if svms.Size() == 0 || volumes.Size() == 0 {
+		return nil
+	}
+
+	topClients, err := v.fetchTopClients(volumes.Values(), svms.Values(), metric)
```
cgrinds, line 306: To reduce unnecessary allocations it would be better to pass the set and iterate over it instead of creating a copy with Values()

cmd/collectors/restperf/plugins/volumetopclients/volumetopclients.go
```diff
@@ -0,0 +1,409 @@
...
+		err    error
+	)
+	if v.volumeInterface != nil {
+		return v.volumeInterface.fetchTopClients(volumes, svms, metric)
+	}
+	query := "api/storage/volumes/*/top-metrics/clients"
+	href := rest.NewHrefBuilder().
+		APIPath(query).
+		Fields([]string{"client_ip", "svm", "volume.name", metric}).
+		Filter([]string{"top_metric=" + metric, "volume=" + strings.Join(volumes, "|"), "svm=" + strings.Join(svms, "|")}).
```
cgrinds, line 401: How likely is it that we hit the Apache request limit with this filter? Worth checking here?

cmd/collectors/restperf/plugins/volumetopclients/volumetopclients.go
```diff
@@ -0,0 +1,409 @@
...
+	mat := v.data[matrixName]
+	if mat == nil {
+		return nil
+	}
+	for i, client := range topClients {
+		clientIP := client.Get("client_ip").String()
+		vol := client.Get("volume.name").String()
+		svm := client.Get("svm.name").String()
+		value := client.Get(metric).Float()
+		instance, err := mat.NewInstance(strconv.Itoa(i))
```
cgrinds, line 320: is `i` stable from poll to poll?

cmd/collectors/restperf/plugins/volumetopclients/volumetopclients.go
```diff
@@ -0,0 +1,409 @@
...
+	topWriteOps := v.getTopN(writeOpsList)
+	topReadData := v.getTopN(readDataList)
+	topWriteData := v.getTopN(writeDataList)
+
+	readOpsVolumes, readOpsSvms := v.extractVolumesAndSvms(data, topReadOps)
+	writeOpsVolumes, writeOpsSvms := v.extractVolumesAndSvms(data, topWriteOps)
+	readDataVolumes, readDataSvms := v.extractVolumesAndSvms(data, topReadData)
+	writeDataVolumes, writeDataSvms := v.extractVolumesAndSvms(data, topWriteData)
+
+	if err := v.processTopClientsByMetric(readOpsVolumes, readOpsSvms, topClientReadOPSMatrix, "iops.read", v.setOpsMetric); err != nil {
```
cgrinds, line 216: setOpsMetric and setDataMetric can be consolidated into a single function if you pass along "ops" or "data" to `processTopClientsByMetric`. That also means you don't need to pass the set function either

grafana/dashboards/cmode/volume.json
```diff
@@ -6457,18 +6457,417 @@
...
+          "gridPos": {
+            "h": 2,
+            "w": 24,
+            "x": 0,
+            "y": 21
+          },
+          "id": 155,
+          "options": {
+            "content": "<br>\n\nTop clients panels are disabled by default and are supported by the Volume Template in RestPerf Collector. For more details on how to enable them, please refer to the discussion at https://github.com/NetApp/harvest/discussions/3130.",
+            "mode": "markdown"
```
Hardikl, line 6485: We can add warning something like: Enabling this may slow down data collection due to a high number of rest calls and post processing for top clients.

Hardikl (commented):
after handling Chris's comment, rest all look good.

cmd/collectors/restperf/plugins/volumetopclients/volumetopclients.go
cgrinds: Let's do one of these:
1. In the current error handling block of `cmd/collectors/restperf/plugins/volumetopclients/volumetopclients.go` at line 393, check if the URL length is > 8KB and if so augment the returned error with that information
2. Change `cmd/collectors/restperf/plugins/volumetopclients/volumetopclients.go` InvokeRestCall to return the following at line 76
`logger.Error().Err(err).Str("href", href).Int("hrefLength", len(href)).Msg("Failed to fetch data")`

PR #3150, opened 2024-09-13
ci: add metric generate docs to CI

rahulguptajss:


cmd/tools/generate/counter.go
```diff
@@ -726,6 +779,10 @@ func generateCounterTemplate(counters map[string]Counter, version [3]int) {
 		panic(err)
 	}
 	fmt.Printf("Harvest metric documentation generated at %s \n", targetPath)
+
+	if issues {
```
cgrinds, line 783: Instead of tracking through a boolean, you could check if the table has any rows
`if table.NumLines() > 0`

Hardikl (dismissed):

jenkins/artifacts/jenkinsfile
```diff
@@ -171,6 +171,18 @@ pipeline {
           }
       }
 
+        stage('Test and Generate Harvest Metrics') {
+            steps {
+                script {
+                    sh 'cp /u/mpeg/harvest/harvest.yml .'
+                    def result = sh(script: 'bin/harvest generate metrics --poller dc1', returnStatus: true)
```
cgrinds, line 178: Can remove the copy if you add `--config /u/mpeg/harvest/harvest.yml`

cgrinds (approved):

PR #3159, opened 2024-09-19
fix: FSA history data should work when multi select

rahulguptajss:


Hardikl (approved):

cgrinds:
Verified in `24.11` ddb97c57
70tas on Discord also verified in an earlier nightly

PR #3172, opened 2024-09-24
perf: reduce allocations when serving Prometheus metrics

cgrinds:


cmd/exporters/prometheus/httpd.go
```diff
@@ -169,35 +173,37 @@ func (p *Prometheus) ServeMetrics(w http.ResponseWriter, r *http.Request) {
...
+// used to keep track of which metrics have been added. The metrics slice is
+// expected to be in the format: # HELP metric_name help text # TYPE metric_name
+// type metric_name{tag="value"} value
+func addMetricsToSlice(data [][]byte, metrics [][]byte, seen map[string]bool, addMetaTags bool) [][]byte {
+
+	if !addMetaTags {
+		return append(data, metrics...)
+	}
+
+	for i, metric := range metrics {
```
Hardikl, line 189: Here, filtering would be on metrics slice, but not on data slice. 
It should be filter data slice as well same as left side ?

Hardikl (approved):

PR #3186, opened 2024-10-01
feat: handle url limit in rest

Hardikl:
This change would be in `build()` function which would be called by rest cli and rest/ems collector calls.


cmd/tools/rest/href_builder.go
```diff
@@ -76,6 +79,10 @@ func (b *HrefBuilder) Build() string {
 
 	href.WriteString("?return_records=true")
 
+	if len(strings.Join(b.fields, ",")) > URLMaxLimit {
```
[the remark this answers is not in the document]
Hardikl: Done

cmd/tools/rest/href_builder.go
```diff
@@ -76,6 +85,26 @@ func (b *HrefBuilder) Build() string {
...
+		fmt.Printf("converting to * due to URL max limit")
```
cgrinds, line 89: we should log nothing or use slog instead of Printf

cmd/tools/rest/href_builder.go
```diff
@@ -76,6 +85,26 @@ func (b *HrefBuilder) Build() string {
...
+		b.fields = []string{"*"}
+	}
+
+	if len(b.hiddenFields) > 0 {
```
cgrinds, line 93: If hidden fields are added below the URLMaxLimit check, they may cause the URL to exceed the max. Hidden fields need to be part of the max check

cmd/tools/rest/href_builder_test.go
```diff
@@ -0,0 +1,101 @@
...
+		ReturnTimeout(&testReturnTimeout).
+		IsIgnoreUnknownFieldsEnabled(isIgnoreUnknownFieldsEnabled).
+		Build()
+
+	if hrefTest1 != expectedHrefTest1 {
+		t.Errorf("hrefTest1 should be %s but got %s", expectedHrefTest1, hrefTest1)
+	}
+
+	testFields = make([]string, 0)
+	for i := range 100000000 {
```
cgrinds, line 31: This takes six seconds to run and isn't really needed, something like this is much faster `	for i := range URLMaxLimit / len("Test") {`

cmd/tools/rest/href_builder_test.go
```diff
@@ -0,0 +1,101 @@
...
+	testReturnTimeout := 10
+	isIgnoreUnknownFieldsEnabled := true
+	expectedHrefTest1 := "api/storage/volumes?return_records=true&fields=name,statistics,svm&return_timeout=10&ignore_unknown_fields=true"
+	hrefTest1 := NewHrefBuilder().
+		APIPath(testQuery).
+		Fields(testFields).
+		HiddenFields(testHiddenFields).
+		Filter(testFilter).
+		ReturnTimeout(&testReturnTimeout).
+		IsIgnoreUnknownFieldsEnabled(isIgnoreUnknownFieldsEnabled).
```
cgrinds, line 23: isIgnoreUnknownFieldsEnabled is always true so this can be changed to
`IsIgnoreUnknownFieldsEnabled(true).`

cmd/tools/rest/href_builder_test.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	expectedHrefTest2 := "api/storage/volumes?return_records=true&fields=*,statistics&return_timeout=10&ignore_unknown_fields=true"
+	hrefTest2 := NewHrefBuilder().
+		APIPath(testQuery).
+		Fields(testFields).
+		HiddenFields(testHiddenFields).
+		Filter(testFilter).
+		ReturnTimeout(&testReturnTimeout).
+		IsIgnoreUnknownFieldsEnabled(isIgnoreUnknownFieldsEnabled).
```
cgrinds, line 42: isIgnoreUnknownFieldsEnabled is always true so this can be changed to
`IsIgnoreUnknownFieldsEnabled(true).`

cmd/tools/rest/href_builder.go
Hardikl: I was thinking that, if hiddenFields would be prior then if it exceed the limit, then it will pass `*` and no hidden fields would be coming in that response. 
But yeah, if it was in later then url rest limit 414 error will come and whole rest call would be failed.

cmd/tools/rest/href_builder_test.go
Hardikl: Done

cmd/tools/rest/href_builder.go
```diff
@@ -76,6 +85,26 @@ func (b *HrefBuilder) Build() string {
...
+			if _, exists := fieldsMap[hiddenField]; !exists {
+				b.fields = append(b.fields, hiddenField)
+				fieldsMap[hiddenField] = true
+			}
+		}
+	}
+
+	if len(strings.Join(b.fields, ",")) > URLMaxLimit {
+		slog.Info("converting to * due to URL max limit")
+		b.fields = []string{"*"}
```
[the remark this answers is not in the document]
Hardikl: Done

PR #3199, opened 2024-10-07
fix: handled flexgroup based on volume config call

Hardikl:


cmd/collectors/restperf/plugins/volume/volume.go
```diff
@@ -31,13 +36,58 @@ func (v *Volume) Init() error {
...
+	return collectors.ProcessFlexGroupData(v.SLogger, data, style, v.includeConstituents, opsKeyPrefix, volumesMap)
+}
+
+func (v *Volume) fetchVolumes() map[string]string {
+	volumesMap := make(map[string]string)
+	query := "api/storage/volumes"
+
+	href := rest.NewHrefBuilder().
+		APIPath(query).
+		Fields([]string{"volume", "volume_style_extended"}).
```
cgrinds, line 69: include `MaxRecords(collectors.DefaultBatchSize).`

cmd/collectors/volume.go
```diff
@@ -33,7 +33,10 @@ func ProcessFlexGroupData(logger *slog.Logger, data *matrix.Matrix, style string
 	cache.UUID += ".Volume"
 
 	for _, i := range data.GetInstances() {
-		if match := flexgroupRegex.FindStringSubmatch(i.GetLabel("volume")); len(match) == 3 {
+		volName := i.GetLabel("volume")
+		switch volumesMap[volName] {
```
cgrinds, line 37: volumesMap may be nil should check at the top and log/return

cgrinds (commented):
Instead of fetching the volumes everytime the performance plugin runs, we should request it at some plugin invocation rate similar to other plugins to reduce the load.

Hardikl:
> Instead of fetching the volumes everytime the performance plugin runs, we should request it at some plugin invocation rate similar to other plugins to reduce the load.

made invocation same as other plugins.

PR #3221, opened 2024-10-23
feat: Adding Ethernet port table from netport template

Hardikl:
image


Hardikl:
As per the REST documentation, MTU unit should be in bytes and not bits/sec, Thanks for raising this.
image

We will correct the unit accordingly.

Hardikl:
Added fix in PR #3278

cgrinds:
Thanks for raising @faguayot This is fixed in the latest nightly

PR #3273, opened 2024-11-08
feat: add tags field in volume table

Hardikl:
image

### DIFF
--- a/conf/rest/9.14.0/volume.yaml
+++ b/conf/rest/9.14.0/volume.yaml
@@ -16,6 +16,7 @@ counters:
   - ^snaplock_type                                => snaplock_type
   - ^snapshot_policy                              => snapshot_policy
   - ^state                                        => state
+  - ^tags                                         => tags
   - ^type                                         => type
   - ^uuid                                         => uuid
   - ^volume_style_extended                        => style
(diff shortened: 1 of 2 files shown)


cmd/tools/grafana/dashboard_test.go
```diff
@@ -699,6 +699,8 @@ func checkVariablesHaveAll(t *testing.T, path string, data []byte) {
 	}
 	exceptionForAllValues := map[string]bool{
 		"cmode/security.json": true,
+		"cmode/cluster.json":  true,
```
cgrinds, line 702: still needed?

pkg/conf/conf_test.go
```diff
@@ -494,6 +494,7 @@ func TestEmbeddedExporter(t *testing.T) {
 	}
 
 	uniqueExporters := GetUniqueExporters(p.Exporters)
+	slices.Sort(uniqueExporters)
```
cgrinds, line 497: no longer needed since GetUniqueExporters sorts the slice

Hardikl: removed.

cmd/tools/grafana/dashboard_test.go
Hardikl: yes it's needed, As we removed `.*` from the `Cluster` var.

cgrinds (approved):

PR #3281, opened 2024-11-11
feat: Harvest should support per-poller prom_ports

cgrinds:
feat: Harvest should support disabled pollers
doc: Explain how prom_port and port range options for the Prometheus exporter work


Hardikl (approved):
looks fine

PR #3303, opened 2024-11-18
perf: reduce the memory footprint of REST collector

cgrinds:
When the REST collector walks pagination links, it does so recursively. Each of these recursive calls retains the previous REST response on the stack, and when there are many recursive calls, a lot of memory can be used. Replace the recursive calls with iterative ones.

### DIFF
--- a/cmd/tools/rest/rest.go
+++ b/cmd/tools/rest/rest.go
@@ -347,39 +347,46 @@ func GetPollerAndAddr(pName string) (*conf.Poller, string, error) {
 
 // FetchForCli used for CLI only
 func FetchForCli(client *Client, href string, records *[]any, downloadAll bool, curls *[]string) error {
-	getRest, err := client.GetRest(href)
-	if err != nil {
-		return fmt.Errorf("error making request %w", err)
-	}
 
-	pollerAuth, err := client.auth.GetPollerAuth()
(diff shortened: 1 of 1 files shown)


cmd/tools/rest/rest.go
```diff
@@ -347,62 +347,69 @@ func GetPollerAndAddr(pName string) (*conf.Poller, string, error) {
...
+		pollerAuth, err := client.auth.GetPollerAuth()
 		if err != nil {
-			return fmt.Errorf("error unmarshalling json %w", err)
+			return err
 		}
-		*records = append(*records, page.Records...)
-	} else {
-		// extract returned records since paginated records need to be merged into a single list
+		*curls = append(*curls, fmt.Sprintf("curl --user %s --insecure '%s%s'", pollerAuth.Username, client.baseURL, href))
+
```
Hardikl, line 365: it should be `nextLink` instead of `href`

Hardikl (approved):

Hardikl (dismissed):

cmd/tools/rest/rest.go
cgrinds: good catch

Hardikl (approved):

PR #3337, opened 2024-11-27
feat: hide transient volumes

Hardikl:


cgrinds:
Let's leave the following comment above all six regexes.

```
# Ignore transient volumes, e.g. SnapProtect, SnapManager, SnapCenter, CommVault, Clone, and Metadata volumes
```

cgrinds (approved):

PR #3345, opened 2024-11-29
doc: keyperf documentation

rahulguptajss:


Hardikl (dismissed):

docs/configure-keyperf.md
```diff
@@ -0,0 +1,137 @@
...
+
+#### Template Example
+
+```yaml
+name:                     Volume
+query:                    api/storage/volumes
+object:                   volume
+
+counters:
+- ^^name                                => volume
```
cgrinds, line 73: Looks like a copy/paste issue. Our templates align the `-` under the `u` in `counters` so this block needs to be shifted to the right by two. Looks to be the case for most sections

cgrinds (approved):

PR #3349, opened 2024-12-02
feat: adding Rwctx template for RestPerf in 9.16.0

Hardikl:
**Fixed in 9.16.0**
image

image

image

-----------------
Cluster is 9.16.1 supporting both Zapi and Rest, Whereas vsim is 9.17.1 and only supporting Rest.
For RestPerf, Qos_xxx counters were not included.

image

For ZapiPerf, 
image

image

image


Hardikl:
> Let's add metric doc as well for this.

Added.

PR #3361, opened 2024-12-06
fix: handle volume cache for performance data

rahulguptajss:


cmd/collectors/restperf/plugins/volume/volume.go
```diff
@@ -64,16 +64,22 @@ func (v *Volume) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix, *util
...
-		// Clean volumesMap map
-		clear(v.volumesMap)
-		v.volumesMap = v.fetchVolumes()
+		// Attempt to fetch new volumes
+		newVolumesMap, err := v.fetchVolumes()
+		if err != nil {
+			v.SLogger.Error("Failed to fetch volumes, retaining cached volumesMap", slog.Any("err", err))
+		} else {
+			// Only update volumesMap if fetchVolumes was successful
+			clear(v.volumesMap)
```
cgrinds, line 73: clear is not needed since you are reassigning in the next line

cgrinds (approved):

PR #3385, opened 2024-12-16
fix: changed var label to ne null from empty

Hardikl:


Hardikl:
Validated against Grafana 11.x.x, other version of Grafana 8, 9, 10 would be pending.

Hardikl:
Validated against Grafana 8.1.2, Grafana 9.5.15, Grafana 10.3.1, Grafana 11.3.1.

Hardikl:
Before the changes,
image



After the changes,
image

PR #3404, opened 2025-01-02
fix: remove redundant label from node template

rahulguptajss:


Hardikl (commented):
uptime is used in this table,
image

PR #3410, opened 2025-01-07
feat: handled empty admin svm in plugin call

Hardikl:


Hardikl:
Testing with CVO is in-progress.

cgrinds (approved):

Hardikl:
It's fail-safe check would be added, not found any case where admin svm is missing in CVO.

Issue #3412, opened 2025-01-07
Support folderUID while importing Grafana Dashbaords


IvanZenger:
**Is your feature request related to a problem? Please describe.**  
I would like to import the Grafana Harvest Dashboards into our Grafana instance under the folder `NetApp/Harvest`. However, when I use the command `--serverfolder "NetApp/Harvest"`, it does not place the dashboards in the subfolder "Harvest" under "NetApp". Instead, it creates a new folder called `NetApp/Harvest` at the root level and imports the dashboards there.

**Describe the solution you'd like**  
Enhance the `grafana import` command with a `--folderuid` flag that allows specifying the unique ID (UID) of the folder where the dashboards should be imported.

**Describe alternatives you've considered**  
Improve the `--serverfolder "NetApp/Harvest"` option to recognize and handle recursive paths correctly.

IvanZenger:
> hi @IvanZenger what version of Grafana are you using? It was only recently that Grafana added support for [subfolders in Grafana 11](https://grafana.com/blog/2024/04/09/grafana-11-release-all-the-new-features/#organize-grafana-dashboards-with-subfolders).
> 
> My read of Grafana's [create folder API](https://grafana.com/docs/grafana/latest/developers/http_api/folder/#create-folder) indicates that for subfolders to work in Grafana you need to enable the `nestedFolders` [feature flag](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/feature-toggles/#preview-feature-toggles) in Grafana and looks like Harvest will need to pass a `parentUid` field when importing subfolders.

Hello @cgrinds,

We are using Grafana version 11.2.3. The FeatureToggle `nestedFolder` is enabled as default since v11.0 and was added as Beta-Feature in v10.0 in Jun 2023.

The parentUid is not required when a dashboard is imported or created. It is only needed if a folder must be created. As in the first proposed solution, using the `--folderUid` parameter, we specify that the folder already exists and does not need to be created. [create/update dashboard](https://grafana.com/docs/grafana/latest/developers/http_api/dashboard/#create--update-dashboard)

In the second solution, the implementation is quite open to interpretation. I would likely parse the path and recursively check if each folder exists. During this process, I would always know the folderUid of the parent folder, which allows me to use it as the parentUid for creating subfolders if they do not exist.

cgrinds:
Thanks @IvanZenger! We'll add this

IvanZenger:
[...]

PR #3417, opened 2025-01-08
feat: Harvest grafana import should support nested Grafana folders

cgrinds:
Thanks to @IvanZenger for reporting

Fixes: #3412


Hardikl (commented):
Do we plan to support in the `grafana export` workflow as well ?

IvanZenger:
fyi: https://github.com/NetApp/harvest/issues/3412#issuecomment-2579259972

cgrinds:
> Do we plan to support in the `grafana export` workflow as well ?

Not required for export

PR #3479, opened 2025-02-17
feat: Volume Audit log

rahulguptajss:


Hardikl (approved):
looks good.
Few comments from UI side,
- We can add few more vars, example volume var and that var should be only consumed in Volume related panels for filtering

cmd/collectors/rest/plugins/auditlog/auditlog.go
```diff
@@ -0,0 +1,252 @@
...
+)
+
+type CacheRefresher interface {
+	RefreshCache() error
+}
+
+var defaultFields = []string{"application", "location", "user", "timestamp", "state"}
+var volumeCache = make(map[string]VolumeInfo)
+var volumeCacheCopy = make(map[string]VolumeInfo)
+var cacheForceRefresh bool
```
cgrinds, line 29: Is force the right name since it seems like this variable prevents refreshes? Maybe a better name is `hasCacheRefreshed` or `hasRefreshedCache`? 

If I understand the life-cycle correctly, this variable will be reset each time `Run()` is called and then called once when the handler's `ExtractNames` runs.

If that's the case, would it make more sense to be more explicit and call RefreshCache() once inside the `if a.HasVolumeConfig()` block in the Run method?

Hardikl (approved):

PR #3498, opened 2025-02-28
fix: workload and workload_volume templates should invoke the instance task before the data task

rahulguptajss:


Hardikl (dismissed):

conf/restperf/default.yaml
```diff
@@ -4,6 +4,7 @@ collector:          RestPerf
 # Order here matters!
 schedule:
   - counter:  24h
+  - instance: 720h   # This is only needed for `workload` and `workload_volume` templates, which overwrite this value in their own templates.
```
cgrinds, line 7: The instance schedule is only used by the `workload` and `workload_volume` templates

cgrinds (approved):

PR #3507, opened 2025-03-05
feat: adding uptime in lif perf templates

Hardikl:
Dashboard changes are in-progress as there are many lif_labels which have svm empty and joining them with lif_uptime may cause issue in table.

### DIFF
--- a/conf/restperf/9.12.0/lif.yaml
+++ b/conf/restperf/9.12.0/lif.yaml
@@ -14,6 +14,7 @@ counters:
   - sent_data
   - sent_errors
   - sent_packets             => sent_packet
+  - up_time                  => uptime
 
 export_options:
   instance_keys:
@@ -21,3 +22,6 @@ export_options:
     - node
     - port
     - svm
+
+override:
+  up_time: raw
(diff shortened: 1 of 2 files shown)


Hardikl:
`SVM` label can be empty in case of Cluster lif and SVM lif.
Those lifs can be identified based on `ipspace` field value, Added one common plugin which would handled them and update `svm` label accordingly when it's empty.

Hardikl:
image

cmd/collectors/lif.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func (l *Lif) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix, *util.Metadata, error) {
+	data := dataMap[l.Object]
+	clusterName := data.GetGlobalLabels()["cluster"]
+
+	for _, lif := range data.GetInstances() {
+		if svm := lif.GetLabel("svm"); svm == "" {
+			if ipspace := lif.GetLabel("ipspace"); ipspace == "Cluster" {
+				svm = ipspace
```
cgrinds, line 24: slightly clearer to change this to `svm = "Cluster"`

PR #3520, opened 2025-03-14
fix: KeyPerf Collector Volume stats are incorrect for flexgroup

rahulguptajss:


Hardikl (approved):

cmd/collectors/rest/rest.go
```diff
@@ -463,7 +470,11 @@ func (r *Rest) ProcessEndPoints(mat *matrix.Matrix, endpointFunc func(e *EndPoin
 			r.Logger.Debug("no instances on cluster", slog.String("APIPath", endpoint.prop.Query))
 			continue
 		}
-		count, _ = r.HandleResults(mat, records, endpoint.prop, true, oldInstances)
+		isEndPoint := true
+		if endpoint.instanceAdd {
+			isEndPoint = false
+		}
+		count, _ = r.HandleResults(mat, records, endpoint.prop, isEndPoint, oldInstances)
```
cgrinds, line 477: Can we clarify this? The endpoint is still an endpoint even if the `instanceAdd` flag is true. What if we change the  `HandleResults` method to call this argument something like `isMutable` or `canCreateOrDelete` or the inverse of these names if the inverse is used more in the `HandleResults` method? 

Alternatively, maybe this would be clearer? `r.HandleResults(mat, records, endpoint.prop, !endpoint.instanceAdd, oldInstances)`

Hardikl (approved):

PR #3524, opened 2025-03-21
fix: EMS cache handling

rahulguptajss:


cmd/collectors/ems/ems.go
```diff
@@ -733,10 +733,12 @@ func (e *Ems) updateMatrix(begin time.Time) {
 		}
 	}
 
-	// remove all ems matrix except parent object
-	mat := e.Matrix[e.Object]
-	e.Matrix = make(map[string]*matrix.Matrix)
-	e.Matrix[e.Object] = mat
+	// clone all present matrix without any data
```
cgrinds, line 736: Initially, I thought this comment meant that we wanted to clone all matrices that do not have data. Maybe it would be clearer to remove the first comment and change the other two lines to

```go
// We want to ensure that the existing matrix is an empty clone so that it gets updated in the Prometheus cache.
// This prevents older instances from appearing in the previous poll.
```

cgrinds (dismissed):

cgrinds (approved):

Hardikl (approved):

PR #3611, opened 2025-05-09
fix: handle HA and volume move alerts

rahulguptajss:


Hardikl (approved):

cmd/collectors/rest/plugins/health/health.go
```diff
@@ -489,12 +493,24 @@ func (h *Health) collectNodeAlerts() int {
 	return nodeAlertCount
 }
 
+func (h *Health) isCustomSystem() bool {
```
cgrinds, line 496: Probably better to move these methods to remote

PR #3613, opened 2025-05-09
feat: update units and comment

Hardikl:
Units were update this way,

Top $TopResources Interface Transmit Traffic -> bytes/sec
Top $TopResources Interface Receive Traffic -> bytes/sec
Top $TopResources Interface Receive Drops -> packets/sec
Top $TopResources Interface Errors -> errors/s
Top $TopResources Interface CRC error -> errors/s

For, switch name and hostname same issue, 
we fetched hostname and saved in `remote.name`, https://github.com/NetApp/harvest/blob/main/cmd/collectors/cisco/rest/client.go#L191
And then assign the value to global var `Switch` here, https://github.com/NetApp/harvest/blob/main/cmd/collectors/cisco/cisco.go#L105
That's why it's always be same.

### DIFF
--- a/conf/keyperf/9.15.0/ethernet_switch_port.yaml
+++ b/conf/keyperf/9.15.0/ethernet_switch_port.yaml
@@ -1,4 +1,4 @@
-
+# This template is in beta, It would be updated later.
 name:               EthernetSwitchPort
 query:              api/network/ethernet/switch/ports
 object:             ethernet_switch_port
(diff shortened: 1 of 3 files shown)


conf/keyperf/9.15.0/ethernet_switch_port.yaml
```diff
@@ -1,4 +1,4 @@
-
+# This template is in beta, It would be updated later.
```
cgrinds, line 1: # This template is in beta, it will be updated later.

conf/rest/9.8.0/ethernet_switch_port.yaml
```diff
@@ -1,4 +1,4 @@
-
+# This template is in beta, It would be updated later.
```
cgrinds, line 1: # This template is in beta, it will be updated later.

PR #3641, opened 2025-05-19
fix: handle uname in qtree zapi plugin

rahulguptajss:


Hardikl (approved):
Good catch @rahulguptajss !!

PR #3642, opened 2025-05-19
doc: fix default schedule values for collector

rahulguptajss:


docs/configure-rest.md
```diff
@@ -310,8 +311,8 @@ of these parameters can be defined in the Harvest or object configuration files
...
+| - `instance`       | duration (Go-syntax)           | poll frequency of updating the instance cache.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |  720 hours |
```
cgrinds, line 315: Since the instance schedule is only used by workload and workload_volume we should mention it here or remove it altogether since it is highlighted in `conf/restperf/default.yaml`

cgrinds (approved):

PR #3661, opened 2025-05-23
feat: Convert EthernetSwitchPort Rest/KeyPerf metrics to cisco_interface when cisco switch 

Hardikl:


Hardikl:
cisco_switch_labels:
image

cisco_interface_admin_up:
image


image


image


image

Hardikl:
Changes highlight:
- Based on ethernet/switch rest api, detect switch is cisco or not
- If cisco switch found
    -> Export metrics with `cisco_xxxxxxxx` way. 
    -> Generate new `cisco_switch_labels` and `cisco_interface_admin_up` metrics which supports in Cisco Dashboard variables.
    -> Few of the panels would be populated with these data.

- If other switch 
     -> Export switch metrics as-is like `ethernet_xxxxxxxx`.

Hardikl:
Added `transmit_packets` and `receive_packets` in networkinterface plugin to match with rest template and updated dashboard json.

Unit test-cases for the changes are in-progress

PR #3672, opened 2025-05-29
feat: handle mcc volumes in rest/zapi based on state and svm

Hardikl:


cmd/collectors/rest/plugins/volume/volume.go
```diff
@@ -179,12 +179,17 @@ func (v *Volume) updateVolumeLabels(data *matrix.Matrix, volumeMap map[string]vo
...
 		vol := volume.GetLabel("volume")
 		tags := volume.GetLabel("tags")
+		volState := volume.GetLabel("state")
 		v.volTagMap[vKey] = volumeTag{vol: vol, svm: svm, tags: tags}
 
 		if !volume.IsExportable() {
 			continue
 		}
 
+		if volState == "offline" && strings.HasSuffix(svm, "-mc") {
```
cgrinds, line 189: Add a comment like
SVM names ending with "-mc" are MetroCluster SVMs. We should only export volume metrics from these SVMs if the volume is online.

PR #3710, opened 2025-06-16
feat: doc includes usage detail of given metric in dashboards

Hardikl:


Hardikl:
updating the `counter.yaml` file to support the dashboard doc for generated metrics.

cmd/tools/generate/counter.go
```diff
@@ -388,25 +420,26 @@ func searchDescriptionSwagger(objName string, ontapCounterName string) string {
...
-func processRestCounters(dir string, client *rest.Client) map[string]Counter {
-	restPerfCounters := visitRestTemplates(filepath.Join(dir, "conf", "restperf"), client, func(path string, client *rest.Client) map[string]Counter {
+func processRestCounters(dir string, client *rest.Client, metricsPanelMap map[string]PanelData) map[string]Counter {
+
+	restPerfCounters := visitRestTemplates(filepath.Join(dir, "conf", "restperf"), client, metricsPanelMap, func(path string, client *rest.Client, metricsPanelMap map[string]PanelData) map[string]Counter {
 		if _, ok := excludePerfTemplates[filepath.Base(path)]; ok {
 			return nil
 		}
-		return processRestPerfCounters(path, client)
+		return processRestPerfCounters(path, client, metricsPanelMap)
```
rahulguptajss: Does it help to make metricsPanelMap as a global map rather than passing it to each method?

cmd/tools/generate/counter.yaml
```diff
@@ -15,12 +19,54 @@ counters:
...
+      - Dashboard: 'ONTAP: Datacenter'
+        Row: Storage Efficiency
+        Panel: Top $TopResources Physical Used with FlexClones by Cluster
 
   - Name: aggr_physical_used_wo_snapshots_flexclones
     Description: Total Data Reduction Physical Used without snapshots and flexclones
+    Panels:
+      - Dashboard: 'ONTAP: Aggregate'
+        Row: Storage Efficiency Ratios
+        Panel: Data Reduction
```
rahulguptajss: How do we get the panel ID for these?

Does it make sense to throw a warning in the logs if such a panel is not found in the dashboard, so that this file is in sync with the dashboards?

cmd/tools/generate/generate.go
```diff
@@ -137,13 +137,26 @@ func doDockerCompose(cmd *cobra.Command, _ []string) {
...
 func doGenerateMetrics(cmd *cobra.Command, _ []string) {
 	addRootOptions(cmd)
-	counters, cluster := BuildMetrics("", "", opts.Poller)
+	metricsPanelMap := make(map[string]PanelData)
+	visitDashboard(
+		[]string{
+			"grafana/dashboards/cmode",
+			"grafana/dashboards/cmode-details",
+			"grafana/dashboards/storagegrid",
+			"grafana/dashboards/cisco",
```
rahulguptajss: Do we need to visit SG and cisco dashboards?

cmd/tools/generate/counter.go
```diff
@@ -243,11 +257,19 @@ func (m MetricDef) TableRow() string {
 	return fmt.Sprintf("| %s | `%s` | `%s` | %s |", m.API, m.Endpoint, m.ONTAPCounter, m.Template)
 }
 
+func (p PanelDef) DashboardTableRow() string {
+	return fmt.Sprintf("| %s | %s | %s | [%s](%s/%s) |", p.Dashboard, p.Row, p.Type, p.Panel, "GRAFANA_HOST", p.PanelLink)
```
cgrinds: Could rewrite as `return fmt.Sprintf("| %s | %s | %s | [%s](GRAFANA_HOST/%s) |", p.Dashboard, p.Row, p.Type, p.Panel, p.PanelLink)`

cmd/tools/generate/counter.go
```diff
@@ -1825,3 +1877,70 @@ func getAllExportedLabels(t *node.Node, counterContents []string) ([]string, []s
...
+	for _, expr := range expressions {
+		allMatches := metricRe.FindAllStringSubmatch(expr.Expr, -1)
+		for _, match := range allMatches {
+			m := match[1]
+			if m == "" {
+				continue
+			}
+
+			key := dashboard + expr.RowTitle + expr.Kind + expr.PanelTitle + expr.PanelID
+			if !slices.Contains(metricsPanelMap[m].Keys, key) {
```
cgrinds: Instead of searching over a slice, would it make more sense to store these in a map so we can do O(1) lookups?

cmd/tools/generate/counter.go
cgrinds: Or turn these function into methods and store the fields in the struct

cmd/tools/generate/counter.yaml
Hardikl: We need to manually update this file. And yes, I would be adding those warnings to sync with dashboard panel and this file.

Hardikl: removed hardcoded values in counter.yaml

cmd/tools/generate/generate.go
```diff
@@ -137,13 +137,26 @@ func doDockerCompose(cmd *cobra.Command, _ []string) {
...
+			visitExpressions(data, metricsPanelMap)
+		})
+	counters, cluster := BuildMetrics("", "", opts.Poller, metricsPanelMap)
 	generateCounterTemplate(counters, cluster.Version)
 }
 
 func doDescription(cmd *cobra.Command, _ []string) {
 	addRootOptions(cmd)
-	counters, _ := BuildMetrics("", "", opts.Poller)
+	metricsPanelMap := make(map[string]PanelData)
```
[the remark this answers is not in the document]
Hardikl: removed this changes

cmd/tools/generate/counter.go
```diff
@@ -1825,3 +1880,71 @@ func getAllExportedLabels(t *node.Node, counterContents []string) ([]string, []s
 	}
 	return metricLabels, append(labels, metricLabels...), isInstanceLabels
 }
+
+func visitDashboard(dirs []string, eachDash func(data []byte)) {
+	for _, dir := range dirs {
+		err := filepath.Walk(dir, func(path string, _ os.FileInfo, err error) error {
+			if strings.Contains(path, "influxdb") {
```
rahulguptajss: Is this check needed if passed directories are of cmode?

cgrinds (approved):

PR #3713, opened 2025-06-18
feat: add filtering support for StatPerf

rahulguptajss:


docs/configure-statperf.md
```diff
@@ -102,4 +102,20 @@ export_options:
...
     - volume
-```
\ No newline at end of file
+```
+
+### Filter
+
+This guide provides instructions on how to use the `filter` feature in StatPerf. Filtering is useful when you need to query a subset of instances. For example, suppose you have a small number of high-value volumes from which you want Harvest to collect performance metrics every five seconds. Collecting data from all volumes at this frequency would be too resource-intensive. Therefore, filtering allows you to create/modify a template that includes only the high-value volumes.
+
+In StatPerf templates, you can set up filters under `counters`. Wildcards like * are useful if you don't want to specify all instances. Please note, ONTAP filtering does not support regular expressions, only wildcard matching with `*`.
```
cgrinds: In StatPerf templates, you can set up filters under the `counters` section using the `filter` key as shown below.

Hardikl (approved):

PR #3715, opened 2025-06-20
doc: add instance_add documentation for endpoints

rahulguptajss:


docs/configure-rest.md
```diff
@@ -197,10 +197,14 @@ The `export_options` section allows you to define how to export these time-serie
...
 
 In Harvest REST templates, `endpoints` are additional queries that enhance the data collected from the main query. The main query, identified by the `query` parameter, is the primary REST API for data collection. For example, the main query for a `disk` object is `api/storage/disks`.
-Typically `endpoints` are used to query the private CLI to add metrics that are not available via ONTAP's public REST API.
+This main query provides the main dataset.
 
+Typically `endpoints` are used to query the private CLI to add metrics that are not available via ONTAP's public REST API.
 Within the `endpoints` section of a Harvest REST template, you can define multiple endpoint entries. Each entry supports its own `query` and associated `counters`, allowing you to collect additional metrics or labels from various API.
-These additional metrics or labels are associated with the main dataset via a key. The key is denoted by the `^^` notation in the counters of both the main query and the `endpoints`.
+These additional metrics or labels are integrated with the main dataset via a key. The key is denoted by the `^^` notation in the counters of both the main query and the `endpoints`.
```
cgrinds: combined instead of integrated?

docs/configure-rest.md
```diff
@@ -197,10 +197,14 @@ The `export_options` section allows you to define how to export these time-serie
...
+
+If the `instance_add` flag is set to `true` within an endpoint, new records will be added rather than amending existing ones.
```
cgrinds: modifying instead of amending? created instead of added?

docs/configure-rest.md
```diff
@@ -197,10 +197,14 @@ The `export_options` section allows you to define how to export these time-serie
 #### Endpoints
 
 In Harvest REST templates, `endpoints` are additional queries that enhance the data collected from the main query. The main query, identified by the `query` parameter, is the primary REST API for data collection. For example, the main query for a `disk` object is `api/storage/disks`.
-Typically `endpoints` are used to query the private CLI to add metrics that are not available via ONTAP's public REST API.
+This main query provides the main dataset.
```
cgrinds: `dataset`, as a concept is not used anywhere else in the documentation. Would it make more sense to stay consistent with Harvest's existing nomenclature and say something like this? "This main query collects disk objects from the ONTAP API and converts them into a matrix". We could link the matrix word to the existing matrix documentation.

If you go with this suggestion, make sure you rewrite all references to dataset on this page.

docs/configure-rest.md
```diff
@@ -267,6 +271,106 @@ export_options:
     - type
 ```
 
+Example with `instance_add`
+
+In the example below, when querying volume data, the main dataset, excluding flexgroups, is collected from `api/storage/volumes`.
```
cgrinds: It might be better to include a few sentences about why this template exists before diving into how `instance_add` is used since it's not obvious why we would query in this way.

Maybe also say this example is from `conf/keyperf/9.15.0/volume.yaml`

docs/configure-rest.md
```diff
@@ -267,6 +271,106 @@ export_options:
...
+  - query: api/private/cli/volume
+    counters:
+      - ^^volume                              => volume
+      - ^^vserver                             => svm
+      - ^aggr_list                            => aggr
+      - ^nodes                                => node
+      - filter:
+          - is_constituent=*
+
+plugins:
```
cgrinds: I don't think we need to copy/paste the rest of the template - it doesn't add anything

PR #3729, opened 2025-06-25
feat: added used% in bucket table in tenant dashboard

Hardikl:
image

### DIFF
--- a/cmd/collectors/storagegrid/plugins/bucket/bucket.go
+++ b/cmd/collectors/storagegrid/plugins/bucket/bucket.go
@@ -5,7 +5,6 @@ import (
 	"github.com/netapp/harvest/v2/cmd/poller/plugin"
 	"github.com/netapp/harvest/v2/pkg/collector"
 	"github.com/netapp/harvest/v2/pkg/conf"
-	"github.com/netapp/harvest/v2/pkg/errs"
 	"github.com/netapp/harvest/v2/pkg/matrix"
 	"github.com/netapp/harvest/v2/pkg/slogx"
 	"github.com/netapp/harvest/v2/third_party/tidwall/gjson"
@@ -29,7 +28,9 @@ func New(p *plugin.AbstractPlugin) plugin.Plugin {
 }
 
 func (b *Bucket) Init(remote conf.Remote) error {
+
 	var err error
+
 	if err := b.InitAbc(); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


cgrinds (approved):

grafana/dashboards/storagegrid/tenant.json
```diff
@@ -758,6 +758,64 @@
...
+              },
+              {
+                "id": "mappings",
+                "value": [
+                  {
+                    "options": {
+                      "match": "null+nan",
+                      "result": {
+                        "index": 0,
+                        "text": "0%"
```
rahulguptajss: Showing 0% is misleading, as this value is not applicable when a quota is not applied to the bucket, meaning its space is not bounded. I think we should either show an empty value or use -, which also works.

cmd/collectors/storagegrid/plugins/bucket/bucket.go
```diff
@@ -66,6 +68,22 @@ func (b *Bucket) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix, *coll
...
 
+	if used = b.data.GetMetric("bytes"); used == nil {
+		return nil, nil, errs.New(errs.ErrNoMetric, "bytes")
+	}
+
+	if quota = b.data.GetMetric("quota_bytes"); quota == nil {
+		return nil, nil, errs.New(errs.ErrNoMetric, "quota_bytes")
+	}
+
+	if usedPercent = b.data.GetMetric("used_percent"); usedPercent == nil {
```
rahulguptajss: Can't we calculate it using the data we have in the dashboard?

grafana/dashboards/storagegrid/tenant.json
Hardikl: changed % logic to grafana and used `-` in case of value are not available.

cmd/collectors/storagegrid/plugins/bucket/bucket.go
```diff
@@ -66,6 +67,14 @@ func (b *Bucket) Run(dataMap map[string]*matrix.Matrix) ([]*matrix.Matrix, *coll
 	// Set all global labels from Rest.go if already not exist
 	b.data.SetGlobalLabels(data.GetGlobalLabels())
 
+	if used = b.data.GetMetric("bytes"); used == nil {
+		return nil, nil, errs.New(errs.ErrNoMetric, "bytes")
+	}
+
+	if quota = b.data.GetMetric("quota_bytes"); quota == nil {
+		return nil, nil, errs.New(errs.ErrNoMetric, "quota_bytes")
+	}
```
rahulguptajss: We don't need this check. quota_bytes may be empty for all buckets, leading to this check being true, which would be incorrect. The dashboard will handle such cases.

PR #3734, opened 2025-06-27
feat: adding legend in dashboard panels

Hardikl:


grafana/dashboards/cmode/power.json
```diff
@@ -701,26 +701,26 @@
       "targets": [
         {
           "exemplar": false,
-          "expr": "sum(environment_sensor_power{cluster=~\"$Cluster\",datacenter=~\"$Datacenter\"})",
+          "expr": "sum by (cluster) (environment_sensor_power{cluster=~\"$Cluster\",datacenter=~\"$Datacenter\"})",
```
rahulguptajss: This doesn't seem correct. We want to show Total power here which is NOT by cluster.
Applicable to other places as well.

Hardikl: image

Regarding this panel in Power dashboard, 
I would say if we always want to show total power in this panel, then the datacenter/cluster filter should not required. Based on filter of datacenter and cluster it doesn't show total power but show total at either cluster level or datacenter level.

Hardikl: reverted this total power case and added panel in exceptionList.

grafana/dashboards/cmode/metadata.json
```diff
@@ -854,10 +854,10 @@
       "targets": [
         {
           "exemplar": false,
-          "expr": "sum(poller_memory_percent{datacenter=~\"$Datacenter\",hostname=~\"$Hostname\",poller=~\"$Poller\"})",
+          "expr": "sum by (hostname) (\n  poller_memory_percent{datacenter=~\"$Datacenter\",hostname=~\"$Hostname\",poller=~\"$Poller\"}\n)",
           "hide": false,
           "interval": "",
-          "legendFormat": "% Memory used",
+          "legendFormat": "% Memory used - {{hostname}}",
```
rahulguptajss, line 860: I think poller would be better for both of these

grafana/dashboards/cmode/network.json
```diff
@@ -3731,38 +3731,38 @@
           "pluginVersion": "8.1.8",
           "targets": [
             {
-              "expr": "sum(fcp_int_count{cluster=~\"$Cluster\",datacenter=~\"$Datacenter\",node=~\"$Node\",port=~\"$FCP\"})",
+              "expr": "sum by (node) (\n  fcp_int_count{cluster=~\"$Cluster\",datacenter=~\"$Datacenter\",node=~\"$Node\",port=~\"$FCP\"}\n)",
```
rahulguptajss: Applicable to others in dashboard.

Adding node legend will make scrolling quite big which is not useful given it already has many metrics in this panel. Keeping these as is better.

grafana/dashboards/cmode/node.json
```diff
@@ -2237,23 +2237,23 @@
           "targets": [
             {
               "exemplar": false,
-              "expr": "avg(node_read_latency{cluster=~\"$Cluster\",datacenter=~\"$Datacenter\",node=~\"$Node\"})",
+              "expr": "avg by (cluster) (node_read_latency{cluster=~\"$Cluster\",datacenter=~\"$Datacenter\",node=~\"$Node\"})",
```
rahulguptajss: Applicable to others in dashboard.

They seem okay as they are. The only use case for aggrgetaion by an object would be if read, write, and others are in their own separate panels.

grafana/dashboards/cmode/svm.json
```diff
@@ -729,25 +729,25 @@
       "targets": [
         {
           "exemplar": false,
-          "expr": "avg(svm_vol_read_latency{cluster=~\"$Cluster\",datacenter=~\"$Datacenter\",svm=~\"$SVM\"})",
+          "expr": "avg by (cluster) (svm_vol_read_latency{cluster=~\"$Cluster\",datacenter=~\"$Datacenter\",svm=~\"$SVM\"})",
```
rahulguptajss: same as earlier comment.

grafana/dashboards/cmode/metadata.json
rahulguptajss: Ignore this one. hostname is correct aggregation here.

cmd/tools/grafana/dashboard_test.go
```diff
@@ -1990,3 +2012,53 @@ func SkipIfMissing(t *testing.T, vars ...string) {
...
+		return true
+	})
+
+	VisitAllPanels(data, func(_ string, _, value gjson.Result) {
+		panelTitle := value.Get("title").ClonedString()
+		if slices.Contains(exceptionList, panelTitle) {
+			return
+		}
+		kind := value.Get("type").ClonedString()
+		if kind == "row" || kind == "table" || kind == "stat" || kind == "bargauge" || kind == "piechart" || kind == "gauge" || kind == "heatmap" {
```
cgrinds: Might be clearer to list what we are looking for instead of what we aren't
```go
if kind != "barchart" && kind != "timeseries" && kind != "text" {
return
}
```

PR #3759, opened 2025-07-11
fix: only show shelves which are local in mcc cluster case

Hardikl:


Hardikl:
Moving state to draft until validated this fix with actual mcc cluster response.

Hardikl:
Validated with local MCC cluster, where 2 shelves are local(1.x) and 2 are remote(2.x).

```
 harvest % ./bin/harvest rest show data -p cluster-mc2 --api storage/shelves --config harvestlatest.yaml | grep local
curl --user admin --insecure 'https://MCC_IP/api/storage/shelves?return_records=true&fields=*' # cluster-mc2
    "local": false,
    "local": false,
    "local": true,
    "local": true,
```

Before this PR fix,
image


image


After the Fix,
image

image

Issue #3802, opened 2025-08-04
cDOT Aggregate dashboard: Possible wrong query and suffix


jowanw:
Version:  harvest-25.08.04-nightly.amd64.deb

<img width="2030" height="603" alt="Image" src="https://github.com/user-attachments/assets/5c8e7533-84d9-4115-9b1e-8afedd3be230" />

Both panels on the right side have a "DATA COMPACTION" suffix, i don't think that's correct.
Bottom right panel "Top x Aggregates by Inactive Data %" is the same query as the one above it, instead of correctly showing % inactive

cgrinds:
Verified on 25.08 branch commit 09c69b8

PR #3803, opened 2025-08-04
fix: Aggregate dashboard should show % inactive

cgrinds:
Fixes: #3802

Thanks @jowanw for reporting!


Hardikl (dismissed):
line 1375 also need to be corrected for suffix

PR #3857, opened 2025-08-25
feat: monitor Snapshot Policy Compliance

rahulguptajss:


integration/test/dashboard_json_test.go
```diff
@@ -73,10 +71,12 @@ var restCounterMap = map[string]struct{}{
 // excludeCounters consists of counters which should be excluded from both Zapi/Rest in CI test
 var excludeCounters = []string{
 	"aggr_physical_",
-	"change_log",
 	"audit_log",
+	"change_log",
 	"cifs_session",
 	"cluster_peer",
+	// Based on the cluster upgrade, these metrics existence would be changed.
```
cgrinds: if we sort this list, this comment does not make sense. It applied to five counters on the left side before. I think it's fine to remove

conf/rest/9.12.0/snapshotpolicy.yaml
```diff
@@ -11,10 +11,10 @@ counters:
   - ^name             => snapshot_policy
   - ^scope            => scope
 
-
 plugins:
+#  - SnapshotViolation
```
cgrinds, line 16: I would have expected `SnapshotViolation` to come after SnapshotPolicy instead of before, both from a chronological and alphabetical point of view

conf/zapi/cdot/9.8.0/snapshotpolicy.yaml
```diff
@@ -20,6 +21,7 @@ plugins:
   - LabelAgent:
       split:
         - policy_owner `-admin` scope,
+#  - SnapshotViolation
```
cgrinds, line 24: I would have expected SnapshotViolation to come after SnapshotPolicy instead of before, both from a chronological and alphabetical point of view

rahulguptajss: SnapshotViolation needs to come first as SnapshotPolicy plugin overwrites copies field which gets used in SnapshotViolation plugin.

cgrinds: We should document that dependency

cgrinds (approved):

cgrinds (approved):

PR #3883, opened 2025-09-10
feat: collect Volume SnapLock information

rahulguptajss:


Hardikl (approved):
changes look good, 
minor comment: if junction_path, style and tags are not required in this new snaplock panel then we don't need to pass in group_left.

PR #3892, opened 2025-09-16
feat: handle spaces in StatPerf

rahulguptajss:


conf/statperf/default.yaml
```diff
@@ -8,4 +8,5 @@ schedule:
 
 objects:
   FlexCache:                flexcache.yaml
-  SystemNode:               system_node.yaml
\ No newline at end of file
+  SystemNode:               system_node.yaml
+  NvmMirror:                nvm_mirror.yaml
```
Hardikl, line 12: default enable may not be required same as iw

rahulguptajss: As discussed, It is not available via RestPerf. So we need to keep it enabled if it is enabled in ZapiPerf.

cmd/collectors/statperf/parser.go
```diff
@@ -182,8 +182,13 @@ func (s *StatPerf) parseInstances(input string) ([]InstanceInfo, error) {
...
 		inst := InstanceInfo{
-			Instance:     strings.TrimSpace(fields[2]),
-			InstanceUUID: strings.TrimSpace(fields[4]),
+			Instance: strings.TrimSpace(fields[2]),
+			// Remove quotes from InstanceUUID when present. UUIDs may be quoted in input data,
+			// but during poll-data calls we receive InstanceUUID without quotes due to the
+			// tabular format, so we remove surrounding quotes here.
+			// Do not modify instance names: they are used as filter values in data-fetch calls
+			// and must retain their quotes.
+			InstanceUUID: strings.Trim(strings.TrimSpace(fields[4]), "\""),
```
cgrinds, line 191: what about using strings.NewReplacer(`"`, "", " ", "")

rahulguptajss: We need to only replace surrounding quotes and spaces. Instance UUIDs don't have any spaces or quotes internally. Instance names can have spaces. The existing Trim covers the use cases we need.

PR #3895, opened 2025-09-17
feat: Harvest MCP

rahulguptajss:


mcp/cmd/server/main.go
```diff
@@ -0,0 +1,1193 @@
...
+		case "INFO":
+			level = slog.LevelInfo
+		case "WARN", "WARNING":
+			level = slog.LevelWarn
+		case "ERROR":
+			level = slog.LevelError
+		}
+	}
+
+	opts := &slog.HandlerOptions{Level: level}
```
cgrinds: handler := slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{
		Level:     level,
		AddSource: true,
		ReplaceAttr: func(_ []string, a slog.Attr) slog.Attr {
			if a.Key == slog.SourceKey {
				source := a.Value.Any().(*slog.Source)
				source.File = filepath.Base(source.File)
			}
			return a
		},
	})

	return slog.New(handler)

mcp/cmd/server/main.go
```diff
@@ -0,0 +1,1193 @@
...
+	handler := slog.NewTextHandler(os.Stderr, opts)
+	return slog.New(handler)
+}
+
+var logger = setupLogger().With(slog.String("component", "mcp-server"))
```
cgrinds: Not sure if we need a "component" key, since all log messages are from the Harvest MCP server. If we want one, perhaps the value should be "harvest-mcp"

mcp/cmd/server/main.go
```diff
@@ -0,0 +1,1193 @@
...
+
+var metricDescriptions map[string]string
+
+// Initialize auth package logger
+func init() {
```
cgrinds: better to not use init unless absolutely necessary. In this case, we don't need it and can move `auth.SetLogger(logger)` to main

mcp/cmd/server/main.go
```diff
@@ -0,0 +1,1193 @@
...
+  HARVEST_TSDB_URL=https://localhost:9090 \
+  HARVEST_TSDB_TLS_INSECURE=true \
+  ./bin/harvest-mcp start
+  
+  # Start with debug logging
+  HARVEST_TSDB_URL=http://localhost:9090 \
+  LOG_LEVEL=DEBUG \
+  ./bin/harvest-mcp start`
+
+type PrometheusResponse struct {
```
cgrinds: better to move these generated types into a type or models file

mcp/cmd/server/main.go
```diff
@@ -0,0 +1,1193 @@
...
+	IncludeDetails bool `json:"includeDetails,omitempty" jsonschema:"Include detailed metrics in the response"`
+}
+
+type ListLabelValuesArgs struct {
+	Label string `json:"label" jsonschema:"Label name to get values for (e.g., 'cluster', 'node', 'volume')"`
+	Match string `json:"match,omitempty" jsonschema:"Optional pattern to filter label values. Supports simple string matching or regex patterns (e.g., '.*prod.*', '^cluster_[0-9]+$')"`
+}
+
+func handlePrometheusError(err error, operation string) *mcp.CallToolResult {
+	logger.Error(operation+" failed", slog.Any("error", err))
```
cgrinds: Import harvest module and replace all `slog.Any("error", err)` calls with `slogx.Err(err)`

mcp/cmd/server/main.go
```diff
@@ -0,0 +1,1193 @@
...
+	return &mcp.CallToolResult{
+		Content: []mcp.Content{
+			&mcp.TextContent{Text: fmt.Sprintf("%s failed: %v", operation, err)},
+		},
+		IsError: true,
+	}
+}
+
+func handleValidationError(message string) *mcp.CallToolResult {
+	return &mcp.CallToolResult{
```
cgrinds, line 161: Do we want logging here too?

mcp/cmd/server/main.go
```diff
@@ -0,0 +1,1193 @@
...
+		}
+	}()
+
+	return io.ReadAll(resp.Body)
+}
+
+func formatDataResponse(data any) (*mcp.CallToolResult, any, error) {
+	content, err := formatJSONResponse(data)
+	if err != nil {
+		return &mcp.CallToolResult{
```
cgrinds, line 189: do we want logging here?

mcp/cmd/server/main.go
```diff
@@ -0,0 +1,1193 @@
...
+
+func addTool[T any](server *mcp.Server, name, description string, handler func(context.Context, *mcp.CallToolRequest, T) (*mcp.CallToolResult, any, error)) {
+	mcp.AddTool(server, &mcp.Tool{
+		Name:        name,
+		Description: description,
+	}, handler)
+}
+
+func executePrometheusQuery(queryURL string, params url.Values) (*PrometheusResponse, error) {
+	config := auth.GetTSDBConfig()
```
cgrinds: instead of recreating the config on every call do we want to read the config once?

mcp/cmd/server/main.go
```diff
@@ -0,0 +1,1193 @@
...
+
+	fullURL := fmt.Sprintf("%s?%s", queryURL, params.Encode())
+	resp, err := auth.MakeRequest(config, fullURL)
+	if err != nil {
+		return nil, fmt.Errorf("request failed: %w", err)
+	}
+	defer func() {
```
cgrinds: What you have is fine, but it's also fine if you use the same pattern we use in the rest of Harvest.
```go
//goland:noinspection GoUnhandledErrorResult
defer resp.Body.Close()
```

mcp/cmd/server/main.go
```diff
@@ -0,0 +1,1193 @@
...
+	if pattern == "" {
+		return items
+	}
+
+	regex, err := regexp.Compile(pattern)
+	if err != nil {
+		logger.Debug("Pattern is not valid regex, using string matching", slog.String("pattern", pattern), slog.Any("error", err))
+		var filtered []string
+		for _, item := range items {
+			if strings.Contains(strings.ToLower(item), strings.ToLower(pattern)) {
```
cgrinds: strings.ToLower(pattern) can be moved outside the for loop

mcp/pkg/auth/auth.go
```diff
@@ -0,0 +1,297 @@
...
+	req, err := http.NewRequest(http.MethodGet, url, http.NoBody)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	if err := addAuthToRequest(req, config.Auth); err != nil {
+		return nil, fmt.Errorf("failed to add auth to request: %w", err)
+	}
+
+	req.Header.Set("User-Agent", "harvest-mcp-server/1.0.0")
```
cgrinds: Use version from version.go

mcp/pkg/rules/files.go
```diff
@@ -0,0 +1,272 @@
...
+	if err := os.WriteFile(tempPath, data, 0600); err != nil {
+		return fmt.Errorf("failed to write temporary file: %w", err)
+	}
+
+	// Rename temporary file to final location
+	if err := os.Rename(tempPath, filePath); err != nil {
+		// Clean up temp file
+		err := os.Remove(tempPath)
+		if err != nil {
+			return err
```
rahulguptajss: This is overwriting original error.

mcp/pkg/rules/files.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+// findRuleInFile searches for a rule by name in the specified file
+func (m *Manager) findRuleInFile(filename, ruleName string) (*AlertRule, error) {
+	ruleFile, err := m.readRuleFile(filename)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, group := range ruleFile.Groups {
```
rahulguptajss, line 208: map would be more efficient for lookup

mcp/pkg/rules/manager.go
```diff
@@ -0,0 +1,341 @@
...
+		rule.Annotations["runbook_url"] = req.NewRunbook
+	}
+
+	// Validate the updated rule
+	if err := rm.validator.ValidateRule(rule); err != nil {
+		return fmt.Errorf("updated rule validation failed: %w", err)
+	}
+
+	// Read the file and update the rule
+	ruleFile, err := rm.fileManager.readRuleFile(filename)
```
rahulguptajss, line 228: We already did that in first line

mcp/pkg/rules/files.go
```diff
@@ -0,0 +1,272 @@
...
+	var ruleFile RuleFile
+	if err := yaml.Unmarshal(data, &ruleFile); err != nil {
+		return nil, fmt.Errorf("failed to parse rule file %s: %w", filename, err)
+	}
+
+	return &ruleFile, nil
+}
+
+// writeRuleFile writes a rule file with backup
+func (m *Manager) writeRuleFile(filename string, ruleFile *RuleFile) error {
```
rahulguptajss, line 96: Does this need to be threadsafe if mcp is hosted as http? Applicable to all file write operations. Probably read also for consistency.

mcp/pkg/rules/files.go
```diff
@@ -0,0 +1,272 @@
...
+
+// determineTargetFile decides which file to use based on rule content
+func (m *Manager) determineTargetFile(ruleName, expression string) string {
+	// Check if this is an EMS-related rule
+	lowerName := strings.ToLower(ruleName)
+	lowerExpr := strings.ToLower(expression)
+
+	if strings.Contains(lowerName, "ems") ||
+		strings.Contains(lowerExpr, "ems_") ||
+		strings.Contains(lowerExpr, "event_") {
```
rahulguptajss: ems events follow ems_events naming. is `event_` correct?

mcp/pkg/rules/prometheus.go
```diff
@@ -0,0 +1,125 @@
...
+	baseURL         string
+	reloadURL       string
+	enableReloadAPI bool
+	client          *http.Client
+	logger          *slog.Logger
+	config          auth.TSDBConfig
+}
+
+// NewPrometheusClient creates a new Prometheus client
+func NewPrometheusClient(config auth.TSDBConfig, logger *slog.Logger) *PrometheusClient {
```
rahulguptajss, line 25: This method can be merged with auth.go createHTTPClient

mcp/pkg/rules/manager.go
```diff
@@ -0,0 +1,341 @@
...
+		return fmt.Errorf("failed to read rule file: %w", err)
+	}
+
+	// Find and remove the rule
+	found := false
+	for groupIdx := range ruleFile.Groups {
+		for ruleIdx, rule := range ruleFile.Groups[groupIdx].Rules {
+			if rule.Alert == req.RuleName {
+				// Remove the rule from the slice
+				ruleFile.Groups[groupIdx].Rules = append(
```
rahulguptajss: Reference of slice removed is still alive?

mcp/pkg/rules/manager.go
cgrinds: The first line returns the alertrule and path, this call returns the ruleFIle

cgrinds (approved):

Hardikl (approved):

PR #3955, opened 2025-10-15
feat: cluster-label flag adds new cluster var/label and update all panels

Hardikl:
…nels while retain existing cluster label/var

Normal import:
image

Import with `--cluster-label netapp_cluster `
image

### DIFF
--- a/cmd/tools/grafana/grafana_test.go
+++ b/cmd/tools/grafana/grafana_test.go
@@ -58,41 +58,12 @@ func TestHttpsAddr(t *testing.T) {
 }
 
 func TestAddPrefixToMetricNames(t *testing.T) {
-
-	var (
-		dashboard                      map[string]any
-		oldExpressions, newExpressions []string
-		updatedData                    []byte
-		err                            error
-	)
-
 	prefix := "xx_"
 	VisitDashboards(
 		[]string{"../../../grafana/dashboards/cisco", "../../../grafana/dashboards/cmode", "../../../grafana/dashboards/cmode-details", "../../../grafana/dashboards/storagegrid"},
 		func(path string, data []byte) {
-			oldExpressions = readExprs(data)
-			if err = json.Unmarshal(data, &dashboard); err != nil {
-				fmt.Printf("error parsing file [%s] %+v\n", path, err)
-				fmt.Println("-------------------------------")
-				fmt.Println(string(data))
-				fmt.Println("-------------------------------")
+			if _, err := expressionCheck(t, data, path, prefix); err != nil {
 				return
-			}
-			addGlobalPrefix(dashboard, prefix)
-
-			if updatedData, err = json.Marshal(dashboard); err != nil {
-				fmt.Printf("error parsing file [%s] %+v\n", path, err)
-				fmt.Println("-------------------------------")
-				fmt.Println(string(updatedData))
-				fmt.Println("-------------------------------")
-				return
-			}
-			newExpressions = readExprs(updatedData)
-
-			for i := range newExpressions {
-				if newExpressions[i] != prefix+oldExpressions[i] {
-					t.Errorf("path: %s \nExpected: [%s]\n     Got: [%s]", path, prefix+oldExpressions[i], newExpressions[i])
-				}
 			}
 		})
 }
@@ -736,6 +707,7 @@ func TestAddClusterLabel(t *testing.T) {
 	}
 
 	newClusterLabel := "net_cluster"
+	varRegex := regexp.MustCompile(`\(([^{]*)\{`)
 	for _, tt := range tests {
 		var (
 			gotVars     []string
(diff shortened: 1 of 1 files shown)


cmd/tools/grafana/grafana.go
```diff
@@ -1558,7 +1631,7 @@ func addImportCustomizeFlags(commands ...*cobra.Command) {
 		cmd.PersistentFlags().BoolVar(&opts.forceImport, "force", false,
 			"Import even if the datasource name is not defined in Grafana")
 		cmd.PersistentFlags().StringVar(&opts.customCluster, "cluster-label", "",
-			"Rewrite all panel expressions to use the specified cluster label instead of the default 'cluster'")
+			"Rewrite all panel expressions to add the specified cluster label and specified cluster variable along with the default 'cluster'")
```
cgrinds, line 1786: Since some folks are already using the `cluster-label` CLI argument to replace the default cluster label. We should create a new CLI argument to add a new variable. Maybe something like `--add-cluster-label` or `--new-cluster-label` That way customers can continue with their previous workflow and opt into this new one with a new CLI arg

Hardikl: Implemented in new cli arg.

cgrinds (dismissed):

cmd/tools/grafana/grafana.go
```diff
@@ -871,6 +904,173 @@ func changeClusterLabel(data []byte, cluster string) []byte {
...
+	// Change all panel expressions
+	VisitAllPanels(newContent, func(path string, _, value gjson.Result) {
+
+		// Rewrite expressions and legends
+		value.Get("targets").ForEach(func(targetKey, target gjson.Result) bool {
+			legendFormat := target.Get("legendFormat")
+			if legendFormat.Exists() {
+				newLegendFormat := rewriteCluster(legendFormat.ClonedString(), cluster)
+				newContent, _ = sjson.SetBytes(newContent, path+".targets."+targetKey.ClonedString()+".legendFormat", []byte(newLegendFormat))
+			}
```
rahulguptajss, line 999: Why do we need to loop twice here? Can it be in done in single loop?

cmd/tools/grafana/grafana.go
```diff
@@ -433,7 +433,36 @@ func newLabelVar(label string) []byte {
   "skipUrlSync": false,
   "sort": 0,
   "type": "query"
-}`, label, cases.Title(language.Und).String(label), label)
+}`, label, title, label)
+}
+
+func newClusterLabelVar(varName string, label string) []byte {
```
rahulguptajss, line 439: If user is using prefix feature in Harvest then this hardcoding of metric name `cluster_new_status` may not work. Could you check once.

cmd/tools/grafana/grafana.go
```diff
@@ -871,6 +904,173 @@ func changeClusterLabel(data []byte, cluster string) []byte {
...
+	templateList := gjson.GetBytes(data, "templating.list")
+	if !templateList.Exists() {
+		fmt.Printf("No template variables found, ignoring add label")
+		return data
+	}
+	vars := templateList.Array()
+
+	// create a new list of vars and copy the existing ones into it, duplicate the fourth var after cluster var since we're going to overwrite it
+	var newVars []gjson.Result
+	newVars = append(newVars, vars[:3]...)
```
rahulguptajss: Code may panic here if variable size is 2. Shouldn't it be dynamic?

cmd/tools/grafana/grafana.go
```diff
@@ -871,6 +904,173 @@ func changeClusterLabel(data []byte, cluster string) []byte {
...
+							newContent, _ = sjson.DeleteBytes(newContent, path+".transformations."+transKey.ClonedString()+".options.indexByName."+clusterN)
+						}
+					}
+
+					// If cluster is excluded from the table, exclude the new cluster label too
+					excludeByName := value.Get("options.excludeByName")
+					if excludeByName.Exists() {
+						clusterIndex := value.Get("options.excludeByName.cluster")
+						if clusterIndex.Exists() {
+							newContent, _ = sjson.SetBytes(newContent, path+".transformations."+transKey.ClonedString()+".options.excludeByName."+cluster, true)
```
rahulguptajss: We should keep value same as earlier for exclude. We are changing it to true always here.

Hardikl: Done.

cmd/tools/grafana/grafana.go
Hardikl: done.

cmd/tools/grafana/grafana.go
Hardikl: Tried both ways, 
```
harvest % bin/grafana import --add-cluster-label netapp_cluster -p  ka
```
and 
```
harvest % bin/grafana import -p ka --add-cluster-label netapp_cluster
```

It's handling prefix for this new var as well
image

cmd/tools/grafana/grafana.go
Hardikl: Done.

cmd/tools/grafana/grafana.go
rahulguptajss: Thanks. Could you add a test case for this so that it doesn't get lost in reordering.

cmd/tools/grafana/grafana.go
rahulguptajss: Thanks. Could you add a test case for this so that it doesn't get lost in reordering.

PR #3958, opened 2025-10-16
fix: handle workload filter batch and instance removal

rahulguptajss:


cmd/collectors/restperf/restperf.go
```diff
@@ -773,15 +790,19 @@ func (r *RestPerf) PollData() (map[string]*matrix.Matrix, error) {
 		}
 	}
 
+	var lastErr error
```
cgrinds: maybe a better name would be noInstanceErr?

cmd/collectors/restperf/restperf.go
```diff
@@ -773,15 +790,19 @@ func (r *RestPerf) PollData() (map[string]*matrix.Matrix, error) {
...
+
```
cgrinds, line 794: A comment mentioning why this change is needed would help. Maybe something like:

When requesting batches, ONTAP sometimes returns no instances for a batch. We need to continue requesting batches when that happens since subsequent batches may still have existing instances. We should not stop batch iteration on errs.ErrNoInstance.

PR #3973, opened 2025-10-27
fix: auditlog dashboard duplicate count in Victoriametrics

rahulguptajss:


cmd/collectors/rest/plugins/auditlog/auditlog.go
```diff
@@ -170,6 +170,8 @@ func (a *AuditLog) constructAuditLogURL(actions []string, state string, timestam
 		Filter([]string{timestampFilter, actionFilter, stateFilter}).
 		Build()
 
+	fmt.Println(href)
```
cgrinds: debug code?

rahulguptajss: removed

PR #4051, opened 2025-12-02
feat: disk cache

cgrinds:


cmd/exporters/prometheus/disk_cache.go
```diff
@@ -38,6 +39,37 @@ type diskCache struct {
...
+	}
+
+	return stats, nil
+}
+
+func (dc *diskCache) exportMetrics(key string, data [][]byte, metricNames *set.Set) {
+	dc.Put(key, data, metricNames)
+}
+
+func (dc *diskCache) streamMetrics(w http.ResponseWriter, _ map[string]struct{}, metrics [][]byte) (int, error) {
```
rahulguptajss, line 61: We need to add locking in this method as well to prevent reading a file that may be in the process of being written by the Export method in prometheus.go

PR #4082, opened 2026-01-09
feat: add units for ONTAP metrics missing them

cgrinds:


cmd/tools/util.go
```diff
@@ -368,9 +368,58 @@ func BuildMetrics(dir, configPath, pollerName string, opts *Options, metricsPane
 		}
 	}
 
+	descriptions := readDescriptions()
+
+	for k, counter := range counters {
+		// Check if there is an overridden description from the descriptions.yaml file and if so use it
+		if desc, ok := descriptions[counter.Name]; ok {
```
rahulguptajss: `counter.yaml` does support description overwrite. Why do we need descriptions.yaml?

cgrinds: good idea. Also helped me find five or six duplicates in counter.yaml

Hardikl (approved):

PR #4088, opened 2026-01-14
feat: ESeries collector infrastructure

rahulguptajss:


cmd/collectors/eseries/plugins/host/host.go
```diff
@@ -0,0 +1,107 @@
...
+				enrichedCount++
+			}
+		}
+	}
+
+	return nil, nil, nil
+}
+
+func (h *Host) buildClusterLookup(systemID string) (map[string]string, error) {
+	clusterNames := make(map[string]string)
```
cgrinds: these 24 lines are the same as volumemapping.go `buildClusterLookup`, maybe refactor into one copy?

cmd/collectors/eseries/rest/client.go
```diff
@@ -0,0 +1,440 @@
...
+		if c.cache.data != nil && !c.cache.expiresAt.IsZero() {
+			newExpiresAt := time.Now().Add(c.cache.ttl)
+			if newExpiresAt.Before(c.cache.expiresAt) {
+				c.cache.expiresAt = newExpiresAt
+			}
+		}
+	}
+
+	// Check if cache is still valid
+	if time.Now().Before(c.cache.expiresAt) && c.cache.data != nil {
```
cgrinds: it's faster to do the nil check first and skip the time math if cache.data is nil

cmd/collectors/eseries/rest/client.go
```diff
@@ -0,0 +1,440 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Populate cache with minimum TTL
+	c.cache.data = data
+	c.cache.expiresAt = time.Now().Add(c.cache.ttl)
+	c.cache.misses++
+
+	c.Logger.Info("cache miss - fetched fresh data",
```
cgrinds: should probably be debug. I don't think most folks care about caching

cmd/collectors/eseries/rest/client.go
```diff
@@ -0,0 +1,440 @@
...
+// Examples:
+//
+//	"11.70.4R1" -> "11.70.4"
+//	"11.90.R4"  -> "11.90.0"
+//	"12.00GA"   -> "12.00.0"
+//	"11.30"     -> "11.30.0"
+//
+// Returns empty string if parsing fails
+func (c *Client) normalizeBundleVersion(bundleDisplay string) string {
+	re := regexp.MustCompile(`^(\d+)\.(\d+)(?:\.(\d+))?`)
```
cgrinds: can be moved out of function so the regex is only compiled once

cmd/collectors/eseries/template.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	if config, ok := configs[objType]; ok {
+		return config
+	}
+	return ObjectConfig{}
+}
+
+func GetESeriesPerfObjectConfig(objType string) ObjectConfig {
+	configs := map[string]ObjectConfig{
+		"controller":  {ArrayPath: "controllerStats", Filter: "type=controller", CalculateUtilization: false, UsesSharedCache: true},
```
cgrinds: minor but since `UsesSharedCache` is always true, you could create a constructor function for arraypath, filter, and CalculateUtilization that defaults UsesSharedCache to true

cmd/collectors/eseries/template.go
```diff
@@ -0,0 +1,203 @@
...
+		return
+	}
+
+	cacheName := ExtractCacheNameFromQuery(e.Prop.Query)
+	if cacheName == "" {
+		return
+	}
+
+	e.Prop.CacheConfig = &rest.CacheConfig{
+		Enabled: true,
```
cgrinds: I don't think the cache is ever disabled so this can probably be removed?

cmd/collectors/eseries/eseries.go
```diff
@@ -0,0 +1,428 @@
...
+
+	credentials := auth.NewCredentials(poller, e.Logger)
+
+	// Use pooled client only if caching is enabled (checked after template parsing)
+	usePool := e.Prop.CacheConfig != nil && e.Prop.CacheConfig.Enabled
+	cacheName := ""
+	if usePool {
+		cacheName = e.Prop.CacheConfig.Name
+	}
+	if e.Client, err = rest.New(poller, duration, credentials, usePool, cacheName); err != nil {
```
cgrinds: do we need to pass both usePool and cacheName - if usePool is false cacheName is "" and if usePool is true, cacheName is not "". Seems like passing only cacheName would be sufficient?

cgrinds (commented):
Nice job! This is looking great.

cmd/collectors/eseries/rest/client.go
```diff
@@ -0,0 +1,439 @@
...
+	if existing, ok := clientPool.Load(key); ok {
+		return existing.(*Client), nil
+	}
+
+	clientPoolMu.Lock()
+	defer clientPoolMu.Unlock()
+
+	if existing, ok := clientPool.Load(key); ok {
+		return existing.(*Client), nil
+	}
```
Hardikl, line 72: Seems to be same as line 67-69?

Hardikl (dismissed):

cmd/collectors/eseries/rest/client.go
```diff
@@ -0,0 +1,435 @@
...
+		slog.String("endpoint", fullPath),
+		slog.Uint64("misses", c.cache.misses),
+		slog.String("ttl", c.cache.ttl.String()),
+		slog.Int("records", len(data)))
+
+	return data, nil
+}
+
+// GetAPIPath returns the API path
+func (c *Client) GetAPIPath() string {
```
cgrinds: minor but there is no need to create a getter, we can use c.ApiPath directly

cmd/collectors/eseries/rest/client.go
```diff
@@ -0,0 +1,435 @@
...
+	DefaultTimeout        = "30s"
+	DefaultAPIPath        = "/devmgr/v2"
+	DefaultESeriesVersion = "11.80.0"
+)
+
+// clientPool stores singleton clients per poller
+var (
+	clientPool   = sync.Map{} // key: pollerName, value: *Client
+	clientPoolMu sync.Mutex
+	re           = regexp.MustCompile(`^(\d+)\.(\d+)(?:\.(\d+))?`)
```
cgrinds: bundleRe might be a better name

cmd/collectors/eseries/eseries.go
```diff
@@ -0,0 +1,427 @@
...
+	if e.Params.HasChildS("labels") {
+		for _, l := range e.Params.GetChildS("labels").GetChildren() {
+			mat.SetGlobalLabel(l.GetNameS(), l.GetContentS())
+		}
+	}
+
+	return nil
+}
+
+func (e *ESeries) InitMisc() error {
```
cgrinds: InitExportOptions instead of InitMisc? Also this method does not return an error

cmd/collectors/eseries/eseries.go
```diff
@@ -0,0 +1,427 @@
...
+	e.Remote = e.Client.Remote()
+
+	return nil
+}
+
+func (e *ESeries) InitMatrix() error {
+	mat := e.Matrix[e.Object]
+	mat.Object = e.Prop.Object
+
+	if e := e.Params.GetChildS("export_options"); e != nil {
```
cgrinds, line 170: maybe e.InitMisc()?

cmd/collectors/eseriesperf/plugins/cachehitratio/cachehitratio.go
```diff
@@ -0,0 +1,182 @@
...
+	cacheHitsIopsTotal := data.GetMetric("cacheHitsIopsTotal")
+	readOps := data.GetMetric("readOps")
+	writeOps := data.GetMetric("writeOps")
+
+	if cacheHitsIopsTotal == nil || readOps == nil || writeOps == nil {
+		c.SLogger.Debug("Missing required metrics for controller/system cache hit ratio")
+		return
+	}
+
+	totalCacheHitRatio, err := data.NewMetricFloat64("totalCacheHitRatio")
```
cgrinds: should be snake instead of camel case

cmd/collectors/eseries/eseries.go
```diff
@@ -0,0 +1,427 @@
...
+	if err := e.Client.Init(1, e.Remote); err != nil {
+		return err
+	}
+
+	e.Remote = e.Client.Remote()
+
+	return nil
+}
+
+func (e *ESeries) InitMatrix() error {
```
cgrinds: this method never returns an error - it does not need a return type

cmd/collectors/eseriesperf/eseriesperf.go
```diff
@@ -0,0 +1,782 @@
...
+	}
+
+	curMat := prevMat.Clone(matrix.With{Data: false, Metrics: true, Instances: true, ExportInstances: true})
+	curMat.Reset()
+
+	systemID := ep.GetCluster()
+
+	// Build query - use filters from template
+	// Note: shared cache may contain broader data, but we still apply template filters
+	filters := ep.Prop.Filter
```
cgrinds, line 295: Could this scenario happen? The cache caches a subset of data due to filtering. A different fetcher using the same cache fetches the data without filters but hits the same cache. Instances are missing since filters were used?

cmd/collectors/eseriesperf/eseriesperf.go
```diff
@@ -0,0 +1,782 @@
...
+			totalUtilMetric.SetValueFloat64(instance, totalUtilization)
+		} else {
+			skips++
+		}
+	}
+
+	return skips, nil
+}
+
+func (ep *EseriesPerf) validateMatrix(prevMat *matrix.Matrix, curMat *matrix.Matrix) error {
```
cgrinds, line 771: including the comment from KeyPerf here would be helpful

cmd/collectors/eseriesperf/eseriesperf.go
rahulguptajss: Filters are disabled for objects that use the cache in the applyFilter method in template.go. I updated the comment to reflect that.

PR #4205, opened 2026-03-25
fix: Cisco collector should ignore zero counters

cgrinds:


cmd/collectors/cisco/plugins/networkinterface/networkinterface.go
```diff
@@ -207,36 +206,19 @@ func (i *Interface) parseInterface(output gjson.Result, envMat *matrix.Matrix) {
...
 
-		inExists := value.Get("eth_inbytes").Exists()
-		outExists := value.Get("eth_outbytes").Exists()
-		if !inExists || !outExists {
-			i.debugMissingBytes(output, interfaceName)
+		spuriousZero := ethClearCounters == "never" && (ethInBytes == 0 || ethOutBytes == 0)
+
+		if spuriousZero {
+			instance.SetExportable(false)
+			i.SLogger.Error(
```
rahulguptajss: Should it be warn level?

PR #4239, opened 2026-04-15
fix: move ssd cache to 12.00

rahulguptajss:


cmd/poller/collector/helpers.go
```diff
@@ -197,20 +197,23 @@ func (c *AbstractCollector) findBestFit(homePath string, confPath string, name s
...
+		originalName[v.String()] = raw
+		versions = append(versions, v)
 	}
 	sort.Sort(version.Collection(versions))
 
 	// get closest index
 	idx := getClosestIndex(versions, ontapVersion)
 	if idx >= 0 && idx < len(versions) {
-		selectedVersion = versions[idx].String()
+		selectedVersion = originalName[versions[idx].String()]
```
cgrinds: Instead of the changes above, can't we use the original version of this code and change this line from `selectedVersion = versions[idx].Original()` to `selectedVersion = versions[idx].Original()`

rahulguptajss: Yes updated.

cgrinds (approved):

Hardikl (approved):

PR #4305, opened 2026-05-27
feat: cmperf collector

rahulguptajss:


cmd/collectors/cmperf/cm2.go
```diff
@@ -0,0 +1,569 @@
...
+		fileSchema   *cmmetrics.ObjectSchema
+		fileSummary  *cmmetrics.CollectionStatus
+		readErr      error
+		schemaLoaded bool
+		metricCount  uint64
+		numPartials  uint64
+	)
+	for rec, msgErr := range cmmetrics.Messages(path) {
+		if msgErr != nil {
+			c.Logger.Warn("error reading CM2 pb file",
```
cgrinds: why warn instead of error?

cmd/collectors/cmperf/cm2.go
```diff
@@ -0,0 +1,569 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("downloadSPIFileONTAP %s: %w", restPath, err)
+	}
+
+	if err := os.WriteFile(destPath, data, 0600); err != nil {
+		return fmt.Errorf("downloadSPIFileONTAP write %s: %w", destPath, err)
+	}
+
+	// TODO validate checksum? checksum URL doesn't seem to be working.
```
cgrinds: we should validate checksum

cmd/collectors/cmperf/cm2.go
```diff
@@ -0,0 +1,569 @@
...
+
+func (c *CmPerf) populateMatrix(oc *cmmetrics.ObjectCollection, curMat *matrix.Matrix) uint64 {
+	schemaMap := c.perfProp.schemaMap
+
+	ts := float64(oc.Timestamp) / 1000.0
+
+	var metricCount uint64
+
+	for _, inst := range oc.Data.Instances {
+		stringVals := make(map[string]string, 8)
```
cgrinds: make(map[string]string) is better unless we're certain that the majority of the time there will be eight values

cmd/collectors/cmperf/cm2.go
```diff
@@ -0,0 +1,569 @@
...
+				slog.String("file", filepath.Base(path)), slogx.Err(msgErr))
+			readErr = msgErr
+			break
+		}
+		if rec.Schema != nil {
+			fileSchema = rec.Schema
+		}
+		if rec.Batch != nil {
+			batch := rec.Batch
+			schema := batch.Schema
```
cgrinds: This should use the fileSchema instead of the batch schema which has been removed

Hardikl (approved):
looks good.

PR #4316, opened 2026-06-10
feat: add volume efficiency template

### LINKED ISSUE #4314: Add built-in Harvest template/config for ONTAP volume efficiency counters
**Is your feature request related to a problem? Please describe.**
We would like Harvest to collect ONTAP volume efficiency information out of the box.
Currently, we are using a custom template/config file under the Harvest `conf` directory to collect volume efficiency counters from:
```
query: api/private/cli/volume/efficiency
object: volume_efficiency
```

This works for our use case, it would be better if these counters were included in the Harvest image/configuration by default so that users can monitor volume efficiency features consistently without custom changes.

**Describe the solution you'd like**
Please add a built-in Harvest template/configuration for ONTAP volume efficiency counters.

Suggested configuration, please include other counters if needed:
```
name:                     VolumeEfficiency
query:                    api/private/cli/volume/efficiency
object:                   volume_efficiency

counters:
  - ^^volume
  - ^^vserver                                     => svm
  - ^compression
  - ^cross_volume_background_dedupe
  - ^cross_volume_inline_dedupe
  - ^data_compaction
  - ^inline_compression
  - ^inline_dedupe
  - ^last_op_state
  - ^path                                         => volume_path
  - ^policy
  - ^state
  - ^storage_efficiency_mode
  - ^type
  - last_op_size
  - progress

export_options:
  instance_keys:
    - svm
    - volume

  instance_labels:
    - compression
    - cross_volume_background_dedupe
    - cross_volume_inline_dedupe
    - data_compaction
    - inline_compression
    - inline_dedupe
    - last_op_state
    - policy
    - state
    - storage_efficiency_mode
    - type
    - volume_path
```
This would allow users to monitor volume efficiency attributes such as compression, inline dedupe, data compaction, storage efficiency mode, policy, state, and last operation details.

**Describe alternatives you've considered**
The current alternative is to maintain a custom template/config file locally under the Harvest `conf` directory.

While this works, it adds operational overhead because the custom configuration must be managed separately from the standard Harvest image and may need to be reapplied or checked after upgrades.

**Additional context**
We are already using the above custom configuration successfully to collect volume efficiency information.

Since these counters are useful for monitoring ONTAP storage efficiency status and configuration at the volume level, it would be ideal to have this template included natively in the NetApp Harvest image/configuration.

Hardikl:
`storage_efficiency_mode` field is not available before 9.10 version in this api.

image

image


conf/rest/default.yaml
```diff
@@ -69,6 +69,7 @@ objects:
   SupportAutoUpdate:           support_auto_update.yaml
   SVM:                         svm.yaml
   Volume:                      volume.yaml
+#  VolumeEfficiency:            vol_efficiency.yaml
```
cgrinds: volume_efficiency.yaml

cgrinds (approved):

PR #4348, opened 2026-07-14
feat: add ssd cache support for older eseries versions

### LINKED ISSUE #4329: Getting SSD Cache details
Our E5700 and E2800 which are on SANtricity 11.90 is using some SSDs as extra readcache.
We cannot see the details of this cache anywhere in the existing graphs, so we suspect they are not harvested?
The only way we know to see some details about this, is to via SANtricity, to to Storage -> Pools & VGs -> Select your SSD Cache Pool -> More -> View SSD Cache Stats...  
This link migh be useful: https://10.0.1.57/devmgr/v2/storage-systems/1/symbol/getFlashCacheStatistics?verboseErrorResponse=true&controller=b
I am told that this is using Symbol API and not collected right now?
I would be very nice to have this detail as a graph... and if possible shown per volume as well, but not even sure if this can be seen anywhere in the current GUI...
We are aware that there is a version 12.00 of SANtricity, but it isn't supported on the E5700/E2800 models or even any E-model... (only seems to be EF-systems)...  and we are not sure if version 12.00 will be released for these "older" systems in the future?   Because there might be another API call for this in version 12 (I am told)...

rahulguptajss:


cmd/collectors/eseries/rest/client_test.go
cgrinds, line 87: Maybe add a test for the new version that was included in client.go 
`//	"12.00.00.9018" -> "12.00.00"`

cmd/collectors/eseriesperf/plugins/controller/controller.go
```diff
@@ -84,6 +87,9 @@ func (c *Controller) refreshControllerLabels(arrayID string) {
 	}
 
 	c.controllerLabels = controllerLabels
+	for uuid, label := range controllerLabels {
```
cgrinds: this map can be built inside the single iteration over the controllers in buildControllerLabelMap

cmd/collectors/eseriesperf/eseriesperf.go
```diff
@@ -97,6 +113,17 @@ func (ep *EseriesPerf) Init(a *collector.AbstractCollector) error {
 		slog.String("timeout", ep.Client.Timeout.String()),
 	)
 
+	if ep.Params.GetChildContentS("type") == "ssd_cache" {
+		ep.useSymbolPath = isLegacyFlashCache(ep)
+		path := "REST"
+		if ep.useSymbolPath {
+			path = "SYMbol"
+		}
+		ep.Logger.Info("ssd_cache fetch path selected",
```
cgrinds, line 118: Since path is only used for logging this could be simplified to something like

```
ep.Logger.Info("ssd_cache fetch path selected",
			slog.Bool("isLegacyFlashCache", ep.useSymbolPath),
			slog.String("version", ep.Remote.Version))
```

cgrinds (approved):

Hardikl (approved):

PR #4377, opened 2026-08-03
feat: add storagegrid_labels metric for StorageGrid

### LINKED ISSUE #4374: [StorageGRID] Add a new metric with `addr` value
Hi,
In the StorageGRID poller, can we add a new _labels metric that includes the `addr` value (from harvest.yml) for the SG instance being polled? 

Something like:
`storagegrid_labels{addr=“10.10.10.10", cluster="RRT-GRID", datacenter="rtp"}`

`addr` value is something that we couldn't add using the template files so requesting this change in the code.

`addr` value is the main request here. Sharing some additional context - a few other labels that we are looking to consume are (a) grid systemId and (b) grid version. 
However, unlike the `addr` field, these two could be added using the template files. 
I'm ONLY listing them here in case it made sense to add these labels to the same above metric since they belong to the same object vs adding separate template files. I'm assuming adding separate template files would generate a new metric instead of appending labels to the above `storagegrid_labels` metric? 

In case considering to include grid systemId and version in the same `storagegrid_labels` metric, I'm attaching example template files
```
name:                       GridLicense
query:                      grid/license
object:                     grid
api:                        v4

counters:
  - ^^systemId              => id

export_options:
  instance_keys:
    - id
  instance_labels:
    - id

``` 

```
name:                       GridVersion
query:                      grid/config/product-version
object:                     grid
api:                        v4

counters:
  - ^^productVersion        => version

export_options:
  instance_keys:
    - version
  instance_labels:
    - version

```

rahulguptajss:
storagegrid_labels{version="12.1.0-20260410.1619.11e43c2",addr="10.10.10.10",system_id="1cf0297f",cluster="GRID",datacenter="rtp"} 1.0

### DIFF
--- a/conf/storagegrid/11.6.0/grid.yaml
+++ b/conf/storagegrid/11.6.0/grid.yaml
@@ -13,6 +13,5 @@ export_options:
   instance_keys:
     - version
   instance_labels:
-    - version
     - addr # set by the Grid plugin
     - system_id # set by the Grid plugin
--- a/conf/storagegrid/11.6.0/grid.yaml
+++ b/conf/storagegrid/11.6.0/grid.yaml
@@ -1,0 +1,18 @@
+name: Grid
+query: grid/config/product-version
+object: storagegrid
+api: v3
+
+counters:
+  - ^^productVersion        => version
+
+plugins:
+  - Grid
+
+export_options:
+  instance_keys:
+    - version
+  instance_labels:
+    - version
+    - addr # set by the Grid plugin
+    - system_id # set by the Grid plugin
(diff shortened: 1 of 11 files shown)


conf/storagegrid/11.6.0/grid.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - ^^productVersion        => version
+
+plugins:
+  - Grid
+
+export_options:
+  instance_keys:
+    - version
+  instance_labels:
+    - version
```
Hardikl: label may not be required after adding as key.

Hardikl (dismissed):

conf/storagegrid/11.6.0/grid.yaml
rahulguptajss: Yes, fixed.

cgrinds (approved):

PR #4385, opened 2026-08-06
feat: add sample-period filtering for cmperf

rahulguptajss:


cmd/collectors/cmperf/cm2.go
```diff
@@ -145,10 +145,12 @@ type cm2FileRecord struct {
 }
 
 func (c *CmPerf) pollONTAPFilesEndpoint(query string) ([]cm2FileRecord, error) {
-	path := "api/cluster/counter-cache/files?fields=path,checksum_path,node,timestamp,object&object=" + url.QueryEscape(query) + "&order_by=timestamp+desc&max_records=1"
+	path := "api/cluster/counter-cache/files?fields=path,checksum_path,node,timestamp,object,sample_period&object=" +
```
cgrinds: we never use the field sample_period so this can be removed

cmd/poller/poller.go
```diff
@@ -1033,14 +1037,19 @@ func buildCmPerfManifest(cols []collector.Collector, manifestName string) []byte
 			continue
 		}
 
+		// At the moment, only the defaultDataPeriod is supported by ONTAP
+		dataPeriod := defaultDataPeriod
+		// Hand the resolved sample period back to the collector so PollData can filter
+		// counter-cache files by the same cadence the manifest requested.
+		if s, ok := col.(samplePeriodSetter); ok {
```
cgrinds: We can remove the interface and this code and replace with
```go
if c, ok := col.(*cmperf.CmPerf); ok {
    c.SetSamplePeriod(dataPeriod)
}
```

Hardikl (dismissed):

Hardikl (dismissed):

cgrinds (approved):

PR #4395, opened 2026-08-07
feat: add clone string to handle memory in cm2

rahulguptajss:


cmd/collectors/cmperf/cmmetrics/metrics.go
```diff
@@ -305,6 +305,16 @@ func readProto(data []byte) (*MetricsFileRecord, error) {
 	return handleMetricsFileRecord(data)
 }
 
+// clonedString clones fc's zero-copy string so it outlives the parse buffer,
+// which lets every struct this package returns be safely retained by callers.
+func clonedString(fc *easyproto.FieldContext) (string, bool) {
+	s, ok := fc.String()
+	if !ok {
+		return s, false
```
cgrinds: return "", false

cgrinds (approved):

PR #4431, opened 2026-08-26
feat: support 1D,2D arrays in cmperf

rahulguptajss:


cgrinds:
LLM review comments


## Findings (priority order)

### 1. High — 2D flatten order is unvalidated and likely disagrees with RestPerf

[populateArrayCounter](cmd/collectors/cmperf/cm2.go) uses `idx := i*len(labelsY) + j` (LabelsX outer, LabelsY fastest):

- Test data assumes `q0/rx, q0/tx, q1/rx, q1/tx`.
- RestPerf `rss_matrix` ([restperf_test.go](cmd/collectors/restperf/restperf_test.go) / [submetrics.json](cmd/collectors/restperf/testdata/submetrics.json)) is the opposite grouping: **all X for Y0, then all X for Y1** (`queue_0#tx_frames, queue_1#tx_frames, …, queue_0#tx_bytes, …`).

If CM2’s packed `List64` follows the same C-array/REST layout, **values will be attached to the wrong (X,Y) pair** while metric names still look plausible. Grafana `metric`/`submetric` labels (`metric=labelX`, `submetric=labelY`) match RestPerf *naming*, but only if the values line up.

The previous `TODO validate against data. wafl_hya_sizer … vsim has it as zero` was removed without a real protobuf fixture. [conf/cmperf/9.19.1/wafl_hya_sizer.yaml](conf/cmperf/9.19.1/wafl_hya_sizer.yaml) only collects `cache_stats`.

**Ask before merge:** dump a real CM2 schema+payload for `rss_matrix` or `wafl_hya_sizer` and assert layout. If it matches RestPerf, change the index to `j*len(labelsX)+i` (or iterate Y outer). Add a golden test from a captured `.pb`, not a hand-written slice.

### 2. Medium — denominator `Exportable` regression (Copilot is right)

In `buildCountersFromSchema`, new code does `m.SetExportable(denomMetric.Exportable)`. Previously (and in RestPerf/StatPerf) denominators were **always** `SetExportable(false)` even when the template already listed the base counter.

If a template includes both a cooked counter and its base, cmperf will start exporting the base. That is extra series, possible dashboard duplication, and a collector inconsistency.

**Fix:** always `m.SetExportable(false)` for denominator metrics created/used as bases (and optionally force `denomMetric.Exportable = false`).

### 3. Medium — supportability of shape mismatches

Per-instance details are `Debug` only. The poll-level Warn is:

- `occurrences` (sum of instance hits)
- `counters` (names only)

On-call will not see expected vs got, 1D vs 2D, or LabelsX/Y sizes unless debug is on. For a new code path this will be the first thing asked in a ticket.

**Fix:** keep one Warn, but include first-mismatch dimensions (e.g. `expected`, `got`, `labelsX`, `labelsY`) or a compact `name=expected/got` list. Avoid logging every instance at Warn.

### 4. Low — array-shaped base forced to `raw`

Forcing `ctrType = "raw"` when the base is array-shaped (including after template `override`) avoids a panic, but exported values will be uncooked cumulatives for `average`/`percent`/`rate` counters. Easy to misread in Grafana.

Prefer skip/not export those numerators, or document clearly in the Warn that values are **not** rates/averages. Also log when an `override` is discarded.

### 5. Low — tests miss the cook path

Unit tests cover metric creation, `HasBaseIndex`, 2D key naming, and mismatch skip. They do **not** cover:

- `cookCounters` on flattened 1D/2D + scalar denominator (delta/divide)
- 2D layout vs a captured CM2 file
- `batchBeforeSchema` drop
- denom `Exportable: false`

`TestBuildCountersFromSchema_CreatesOnCurMat` does not assert `counterInfo["read_io_type"].denominator == "read_io_type_base"`.

### 6. Nit — dead branch

Inside `if len(cs.LabelsY) > 0`, `len(labelsY) == 0` cannot happen. Harmless; drop it.

cmd/collectors/cmperf/cm2.go
```diff
@@ -139,42 +125,73 @@ func hasHistogramSuffix(name string) bool {
...
+			}
+		}
 
 		if ov := c.GetOverride(name); ov != "" {
 			ctrType = ov
 		}
+		if baseIsArrayShaped {
+			// Per-label division isn't possible regardless of what the template overrides to.
+			if inTemplate {
+				if ov := c.GetOverride(name); ov != "" && ov != "raw" {
```
cgrinds: reuse c.GetOverride(name) from line 165

cmd/collectors/cmperf/cm2.go
```diff
@@ -626,27 +664,32 @@ func (c *CmPerf) populateArrayCounter(
...
 				idx := i*len(labelsY) + j
-				if idx >= len(values) {
-					c.Logger.Warn("2D array counter: values exhausted before labels",
-						slog.String("counter", cs.Name),
-						slog.Int("expected", len(labelsX)*len(labelsY)),
-						slog.Int("got", len(values)),
-					)
-					return count
-				}
 				k := cs.Name + arrayKeyToken + labelX + arrayKeyToken + labelY
```
cgrinds: key is fully re-concatenated per cell per instance, move the cs.Name + "#" + labelX prefix to the outer loop

cmd/collectors/cmperf/cm2.go
```diff
@@ -139,42 +125,73 @@ func hasHistogramSuffix(name string) bool {
...
+			// Per-label division isn't possible regardless of what the template overrides to.
+			if inTemplate {
+				if ov := c.GetOverride(name); ov != "" && ov != "raw" {
+					c.Logger.Warn("template override discarded, base counter is array-shaped",
+						slog.String("counter", name),
+						slog.String("override", ov),
+					)
+				}
+			}
+			ctrType = "raw"
```
cgrinds, line 181: Does this leave the metric exportable, so a CookAverage/CookPercent counter ships its raw cumulative ONTAP value under a name customers read as a latency/percent? Might be better to not export rather than publish wrong units

rahulguptajss:
> LLM review comments
> 
> ## Findings (priority order)
> ### 1. High — 2D flatten order is unvalidated and likely disagrees with RestPerf
> [populateArrayCounter](cmd/collectors/cmperf/cm2.go) uses `idx := i*len(labelsY) + j` (LabelsX outer, LabelsY fastest):
> 
> * Test data assumes `q0/rx, q0/tx, q1/rx, q1/tx`.
> * RestPerf `rss_matrix` ([restperf_test.go](cmd/collectors/restperf/restperf_test.go) / [submetrics.json](cmd/collectors/restperf/testdata/submetrics.json)) is the opposite grouping: **all X for Y0, then all X for Y1** (`queue_0#tx_frames, queue_1#tx_frames, …, queue_0#tx_bytes, …`).
> 
> If CM2’s packed `List64` follows the same C-array/REST layout, **values will be attached to the wrong (X,Y) pair** while metric names still look plausible. Grafana `metric`/`submetric` labels (`metric=labelX`, `submetric=labelY`) match RestPerf _naming_, but only if the values line up.
> 
> The previous `TODO validate against data. wafl_hya_sizer … vsim has it as zero` was removed without a real protobuf fixture. [conf/cmperf/9.19.1/wafl_hya_sizer.yaml](conf/cmperf/9.19.1/wafl_hya_sizer.yaml) only collects `cache_stats`.
> 
> **Ask before merge:** dump a real CM2 schema+payload for `rss_matrix` or `wafl_hya_sizer` and assert layout. If it matches RestPerf, change the index to `j*len(labelsX)+i` (or iterate Y outer). Add a golden test from a captured `.pb`, not a hand-written slice.
> 
> ### 2. Medium — denominator `Exportable` regression (Copilot is right)
> In `buildCountersFromSchema`, new code does `m.SetExportable(denomMetric.Exportable)`. Previously (and in RestPerf/StatPerf) denominators were **always** `SetExportable(false)` even when the template already listed the base counter.
> 
> If a template includes both a cooked counter and its base, cmperf will start exporting the base. That is extra series, possible dashboard duplication, and a collector inconsistency.
> 
> **Fix:** always `m.SetExportable(false)` for denominator metrics created/used as bases (and optionally force `denomMetric.Exportable = false`).
> 
> ### 3. Medium — supportability of shape mismatches
> Per-instance details are `Debug` only. The poll-level Warn is:
> 
> * `occurrences` (sum of instance hits)
> * `counters` (names only)
> 
> On-call will not see expected vs got, 1D vs 2D, or LabelsX/Y sizes unless debug is on. For a new code path this will be the first thing asked in a ticket.
> 
> **Fix:** keep one Warn, but include first-mismatch dimensions (e.g. `expected`, `got`, `labelsX`, `labelsY`) or a compact `name=expected/got` list. Avoid logging every instance at Warn.
> 
> ### 4. Low — array-shaped base forced to `raw`
> Forcing `ctrType = "raw"` when the base is array-shaped (including after template `override`) avoids a panic, but exported values will be uncooked cumulatives for `average`/`percent`/`rate` counters. Easy to misread in Grafana.
> 
> Prefer skip/not export those numerators, or document clearly in the Warn that values are **not** rates/averages. Also log when an `override` is discarded.
> 
> ### 5. Low — tests miss the cook path
> Unit tests cover metric creation, `HasBaseIndex`, 2D key naming, and mismatch skip. They do **not** cover:
> 
> * `cookCounters` on flattened 1D/2D + scalar denominator (delta/divide)
> * 2D layout vs a captured CM2 file
> * `batchBeforeSchema` drop
> * denom `Exportable: false`
> 
> `TestBuildCountersFromSchema_CreatesOnCurMat` does not assert `counterInfo["read_io_type"].denominator == "read_io_type_base"`.
> 
> ### 6. Nit — dead branch
> Inside `if len(cs.LabelsY) > 0`, `len(labelsY) == 0` cannot happen. Harmless; drop it.

First 2 comments from LLM are incorrect. Updated PR with other logging and test related comments.

cmd/collectors/cmperf/cm2.go
rahulguptajss: Added

cgrinds (approved):

PR #4433, opened 2026-08-26
fix: ssdcache drive,volume should parse arrays

### LINKED ISSUE #4432: E-Series array-labelhandling breaks SSD cache and volume mapping labels
PR https://github.com/NetApp/harvest/pull/4413 changed how array-valued ^ labels are serialized in eseries.go — from raw JSON array text to a sorted, comma-joined string. This was needed for the new controller ip address labels, but it's shared code used by every array-typed label across E-Series, and it silently broke two plugins that still expected the old JSON format:

ssdcachecapacity.go — cached_volume_ids/drive_ids labels were parsed with gjson.Parse expecting JSON; with the new comma format this misparses and every SSD cache shows 0 cached volume/drive.
volumemapping.go — list_of_mappings/metadata labels (arrays of objects) hit the same issue; luns, hosts, mapping_types, and workload labels have never been populated since this regression.
Separately, an unrelated pre-existing bug in eseriesperf.go was found during the same investigation: the SSD cache ssd_cache name global label is set on a matrix instance that's discarded before export, so it's always missing from perf metrics.

rahulguptajss:


cmd/collectors/eseries/plugins/volumemapping/volumemapping.go
```diff
@@ -216,11 +216,12 @@ func (v *VolumeMapping) buildWorkloadLookup(systemID string) (map[string]string,
...
-	if listOfMappingsJSON == "" || listOfMappingsJSON == "[]" {
+	if listOfMappingsJSON == "" {
 		return
 	}
 
-	mappings := gjson.Parse(listOfMappingsJSON)
+	// eseries.go serializes array-type labels as comma-joined object blobs, not a JSON array; rewrap to parse.
+	mappings := gjson.Parse("[" + listOfMappingsJSON + "]")
 	if !mappings.IsArray() {
 		return
```
cgrinds, line 229: Maybe log here so we don't silently fail again?

cmd/collectors/eseries/plugins/volumemapping/volumemapping.go
```diff
@@ -291,11 +292,12 @@ func (v *VolumeMapping) addPoolLabel(volumeInstance *matrix.Instance, poolNames
...
-	if metadataJSON == "" || metadataJSON == "[]" {
+	if metadataJSON == "" {
 		return
 	}
 
-	metadata := gjson.Parse(metadataJSON)
+	// eseries.go serializes array-type labels as comma-joined object blobs, not a JSON array; rewrap to parse.
+	metadata := gjson.Parse("[" + metadataJSON + "]")
 	if !metadata.IsArray() {
 		return
```
cgrinds, line 308: Maybe log here so we don't silently fail again?

cgrinds (commented):
Maybe add a comment around cmd/collectors/eseries/eseries.go:376 that the join produces two different shapes that plugins need to handle: split on comma or wrap with [ ] then parse JSON.

My concern is the next plugin copy/pastes the wrong pattern

cmd/collectors/eseries/plugins/volumemapping/volumemapping.go
rahulguptajss: Added.

cgrinds (approved):

PR #4474, opened 2026-09-21
fix: EMS should bound its collection window

cgrinds:


conf/ems/default.yaml
```diff
@@ -1,6 +1,8 @@
 collector: Ems
 
-client_timeout: 1m
+#  Keep this below the data interval below.
+client_timeout: 2m
```
rahulguptajss: Add handling in the collector to prevent it from starting if this condition is not met?

rahulguptajss: Is client_timeout read correctly by collecor?

cmd/collectors/ems/ems.go
```diff
@@ -267,18 +383,228 @@ func (e *Ems) getTimeStampFilter(clusterTime time.Time) string {
...
+		// allowance, so several hrefs cannot together collect a multiple of
+		// max_records.
+		//
+		// The time check refuses to *begin* a page unless a whole
+		// client_timeout still fits inside the budget. The REST client takes
+		// no context, so a request already in flight cannot be cut short;
+		// without this a page starting just before the deadline could overrun
+		// the poll interval by the full timeout, which is the pile-up the
+		// budget exists to prevent.
+		if len(records) >= quota || time.Until(deadline) < e.Client.GetTimeout() {
```
rahulguptajss: With client_timeout set to 2m and data set to 3m, does that mean only one page call is allowed?

cmd/collectors/ems/ems.go
```diff
@@ -267,18 +381,255 @@ func (e *Ems) getTimeStampFilter(clusterTime time.Time) string {
...
+
+	err := rest.FetchAllStream(e.Client, &e.RequestMetadata, href, func(batch []gjson.Result, _ int64) error {
+		for _, record := range batch {
+			// Client-side end bound. This endpoint rejects a two-sided filter
+			// on `time`, so the end of the window is enforced here - see
+			// timeWindow.
+			if !inWindow(record, toTime) {
+				continue
+			}
+			records = append(records, record)
```
rahulguptajss: If records were higher than quota then they are already appended here and not truncated?

cmd/collectors/ems/ems.go
```diff
@@ -267,18 +381,255 @@ func (e *Ems) getTimeStampFilter(clusterTime time.Time) string {
...
+// each in turn. The event names are split across hrefs in a stable order, so a
+// first-come allowance lets one flooding event name consume the whole cap and
+// leave the trailing names unqueried - permanently, since the split does not
+// change between polls. A rare LUN.offline in the last chunk would go
+// uncollected because wafl.vol.autoSize.done is noisy in the first.
+//
+// Unused allowance rolls forward: an href that returns less than its share
+// leaves the remainder to those after it, so dividing the cap costs nothing
+// when no href is flooding.
+func hrefQuota(maxRecords, taken, i, hrefs int) int {
```
rahulguptajss: A higher even count, like 300+, may then miss events because it may not get the quota. Shouldn't each href get a fresh maximum number of records?

cmd/collectors/ems/window_test.go
```diff
@@ -0,0 +1,447 @@
...
+
+	for i, q := range quotas {
+		if q <= 0 {
+			t.Errorf("href %d got quota %d, every href must be able to ask for records", i, q)
+		}
+	}
+	assert.True(t, taken <= maxRecords)
+}
+
+// client_timeout defaults to 2m against a 3m data interval. A poller with a
```
rahulguptajss: default is 1m in default.yaml

cmd/collectors/ems/ems.go
```diff
@@ -267,18 +381,255 @@ func (e *Ems) getTimeStampFilter(clusterTime time.Time) string {
...
+// collect walks every href within a shared record cap and time budget.
+func (e *Ems) collect(hrefs []string, deadline time.Time, toTime int64) ([]gjson.Result, bool, error) {
+	var (
+		records   []gjson.Result
+		truncated bool
+		skipped   int
+	)
+
+	for i, h := range hrefs {
+		quota := hrefQuota(e.maxRecords, len(records), i, len(hrefs))
```
Hardikl: Need to better understand quota, does based on quota records(available or filled) we reduced not to export/process all or any records?

cmd/collectors/ems/ems.go
```diff
@@ -426,6 +436,23 @@ func sortRecords(records []gjson.Result) {
...
+		// `time`, so the end of the window is enforced here - see timeWindow.
+		if !inWindow(record, toTime) {
+			continue
+		}
+		if len(records) >= quota {
+			return records, true
+		}
+		records = append(records, record)
+	}
+	return records, false
```
rahulguptajss: If the record size is 25000 and the batch size is 10000, are we going to hit false and query one more time? We should stop if the maximum record count is reached otherwise, there will be an extra, unnecessary call.
So, maybe we should append first and then check the size?

cmd/collectors/ems/ems.go
```diff
@@ -267,18 +369,233 @@ func (e *Ems) getTimeStampFilter(clusterTime time.Time) string {
...
+// is left to begin another page. It never escapes fetchEMSData.
+var errWalkBudget = errors.New("ems walk budget exhausted")
+
+// fetchEMSData pages through href, dropping records past the end of the window
+// and stopping once too little of the time budget is left to begin another
+// page. The bool reports that it stopped for that reason.
+//
+// Partial results are returned rather than discarded. The previous behavior
+// threw away every record already retrieved when a walk failed, so the most
+// expensive polls were also the ones that produced nothing.
```
rahulguptajss, line 475: Previous behaviour (in main) didn't do that?

rahulguptajss (approved):

cmd/collectors/ems/ems.go
cgrinds: In main, a failed page unwinds through four layers that each return nil:

1. fetchAll returns the error on the failing page
2. FetchAll does return nil, err — the pages it had already accumulated are dropped
3. GetRestData → handleError returns nil, fmt.Errorf(...)
4. fetchEMSData and PollData both return nil, err

e.g. 31 pages fetched, page 32 times out, nothing is exported.

cgrinds: that said, including previous behavior comments is not helpful so I'm removing

